PR #1, opened 2017-10-10
Migrated existing validation logic

jonnyshaw89:
This PR moved the Schema validation logic out of https://github.com/ONSdigital/eq-survey-runner and into its own repo that exposes a Rest interface as well as being able to be imported and run against local files.


Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM python:3.6-onbuild
```
jcox-dev: Perhaps add a `.python-version` file as well for virtualenvs.

api/validate.py
```diff
@@ -0,0 +1,30 @@
...
+
+validate_blueprint = Blueprint('validate', __name__)
+
+validator = Validator()
+
+
+@validate_blueprint.route('/validate', methods=['POST'])
+def validate_schema():
+    json_to_validate = None
+
```
jcox-dev: We should probably handle the standard exceptions that are likely to be raised:

```python
...
import http.client
...
    try:
        if 'url' in values:
            with urllib.request.urlopen(values['url']) as url:
                json_to_validate = json.loads(url.read().decode())        

        if request.data:
            json_to_validate = json.loads(request.data.decode())

        errors = validator.validate_schema(json_to_validate)

    except (ValueError, json.decoder.JSONDecodeError) as e:
        errors = [f'JSON decode error: {e}']

    except (http.client.error, urllib.error.HTTPError) as e:
        errors = [f'HTTP error: {e}']

    except (TypeError, KeyError) as e:
        errors = [f'Schema parse error: {e}']
...
```

jcox-dev: Unless it should just `flask.abort(417)` instead...?

api.py
```diff
@@ -0,0 +1,15 @@
+from flask import Flask
+from flask_script import Manager
+from flask_script import Server
```
ajmaddaford: We should review whether using flask script is needed, https://flask-script.readthedocs.io/en/latest/ recommends looking at flasks built in cli tool.

api/status.py
```diff
@@ -0,0 +1,8 @@
+from flask import Blueprint, Response
+
+status_blueprint = Blueprint('status', __name__)
+
+
+@status_blueprint.route('/status', methods=['GET'])
+def validate_schema():
+    return Response(status=200)
```
ajmaddaford: It's probably worth adding the application version to the status response as we do in https://github.com/ONSdigital/eq-survey-runner/blob/master/app/setup.py#L265

api/validate.py
```diff
@@ -0,0 +1,39 @@
...
+validator = Validator()
+
+
+@validate_blueprint.route('/validate', methods=['POST'])
+def validate_schema():
+    data = None
+
+    values = request.form
+    if 'url' in values:
```
ajmaddaford: It would be better if the url functionality worked by a GET request `url` query param. It would allow quick testing of a schema via a simple link rather than having to construct a POST request.

schema/schema_v1.json
```diff
@@ -0,0 +1,675 @@
...
+  "eq_id": {
+    "type": "string",
+    "description": "Used in combination with the form_type to uniquely identify a questionnaire.",
+    "required": "false"
+  },
+  "form_type": {
+    "type": "string",
+    "description": "Used in combination with the eq_id to uniquely identify a questionnaire.",
+    "required": "false"
+  },
```
ajmaddaford: `eq_id` and `form_type` were moved in the eq-survey-runner pull request https://github.com/ONSdigital/eq-survey-runner/pull/1298. Do we have a story to remove the schemas from survey runner and run this as a service in the build?

requirements_for_test.txt
```diff
@@ -0,0 +1 @@
+pytest==2.8.2
```
ajmaddaford: Minor, I think the file should be named `test_requirements.txt`, otherwise tab complete stops before `.txt` when installing the application dependencies from `requirements.txt`

tests/test_schema_id_regex.py
```diff
@@ -0,0 +1,126 @@
...
+
+from jsonschema import ValidationError, validate
+
+def create_schema_with_id(schema_id='answer'):
+    """
+    Utility method that loads a JSON schema file and swaps out an answer Id.
+    :param schema_id: The Id to use for the answer.
+    :return: The JSON file with the Id swapped for schema_id
+    """
+    schema_path = "tests/schemas/test_percentage.json"
```
ajmaddaford: We should rename the schema to `test_schema_id_regex.json` and simplify it to it's minimum - removing navigation and empty properties.

jonnyshaw89:
@ajmaddaford with the flask CLI the port is passed in when running it, thus splitting the app from how it is run.

jonnyshaw89:
@ajmaddaford the JSON output pretty printed as we are using jsonify?
https://github.com/ONSdigital/eq-schema-validator/pull/1/files#diff-6552e6bc6d7384fc78b3725037abe9a0R39

ajmaddaford:
@jonnyshaw89 
- The port should be configurable by an env var, so that someone can run `./scripts/run_app.sh` on a custom port without modifying it. Something like:
  ```
  port=${PORT:-5001}
  flask run --port $port
  ```
- Following on from our discussion about the JSON schema errors not being pretty printed, we should separate the JSON schema validation from the extended validation. If JSON schema validation fails we should return immediately.

Dockerfile
jcox-dev: `-onbuild` images are now considered [deprecated](https://github.com/docker-library/official-images/issues/2076).

requirements.txt
```diff
@@ -0,0 +1,4 @@
+flask==0.12.2
```
jcox-dev: Probably one for another card but as it's a new project, let's make sure we're using [Pipenv](https://docs.pipenv.org/index.html).

app/validation/validator.py
```diff
@@ -48,9 +47,9 @@ def _validate_json_against_schema(self, json_to_validate):
             validate(json_to_validate, self.schema, resolver=resolver)
             return []
         except ValidationError as e:
-            return ['Schema Validation Error! does not validate against schema. Error [{}]'.format(e)]
+            return ['Message [{}] Path [{}]'.format(e.message, e.path)]
```
ajmaddaford: I think it would be worth returning this in a structured format, something like:
```
{ 
  'message': e.message,
  'path': e.path
}
```

app/validation/validator.py
```diff
@@ -48,9 +47,9 @@ def _validate_json_against_schema(self, json_to_validate):
...
         except SchemaError as e:
-            return ['JSON Parse Error! Error [{}]'.format(e)]
+            return ['{}'.format(e)]
```
ajmaddaford: If we're only ever returning one error we probably don't need to be wrapping it in an array here

ajmaddaford:
I think we need to make the error response more consistent. It should always return an errors element that is an array of objects, so schema errors would be:
```
{
  "errors": [{
    "message": "'group_order' is a required property",
    "path": "deque(['navigation', 'sections', 13])"
  }]
}
```
and other errors would be:
```
{
  "errors": [
    { 
      "message": "Schema Integrity Error. Routing rule routes to invalid block [years-internal-investment-r-d2]" 
    },
    { 
      "message": "Schema Integrity Error. Routing rule routes to invalid block [investment-purposes-innovation2]"
    }
  ]
}
```

ajmaddaford:
I think it should return `{}` rather than `{ "errors": [] }` when there are no errors

Issue #2, opened 2017-10-17
Use Pipenv / Pipfile for dependency management


jcox-dev:
Details about [how to import from a requirements](https://docs.pipenv.org/basics.html#importing-from-requirements-txt) file. (NB: would want the form `package_name = "*"` in Pipefile.)

PR #3, opened 2017-10-25
Replace requirements.txt with Pipfile and lock for use with Pipenv

jcox-dev:
Change to using Pipenv for dependencies and virtual environment management.

Setup: `pip install -U pip pipenv`
Run: `pipenv run scripts/run_app.sh`

Fixes: #2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM python:3.6
 
+RUN pip install pipenv==8.2.7
+
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM python:3.6
```
shahi645: Should be 3.4.2 to keep inline with Survey Runner? Or is there a reason we need 3.6?

shahi645:
@jcox-dev  
Should we integrate a README as part of this PR?

jcox-dev:
@shahi645 it would be good to have but I think a separate PR for that would make more sense. Maybe open a new issue?

Dockerfile
```diff
@@ -1,5 +1,7 @@
...
 
+RUN pip install -U pipenv==8.2.7
```
jonnyshaw89: I think we should remove -U as it contradicts the pinned version? 
Thats probably why codacy is moaning.

PR #5, opened 2017-11-14
Improve JSON schema validation for types

jonnyshaw89:
I have split the schema across multiple files.
Each type of block/question/answer not has its own definition in its own file.

Does it still validate schemas?


schema/answers/checkbox.json
```diff
@@ -0,0 +1,84 @@
...
+      },
+      "guidance": {
+        "$ref": "../common_definitions.json#/guidance"
+      },
+      "description": {
+        "type": "string"
+      },
+      "type": {
+        "type": "string",
+        "pattern": "^Checkbox$"
```
ajmaddaford: Did you consider using an enum of one value for types rather than a regex? I think `"enum": ["Checkbox"]` reads better as it's on a single line and there's no regex characters.

schema/answers/relationship.json
```diff
@@ -0,0 +1,76 @@
...
+            "label": {
+              "type": "string"
+            },
+            "value": {
+              "type": "string"
+            },
+            "q_code": {
+              "$ref": "../common_definitions.json#/q_code"
+            },
+            "child_answer_id": {
```
ajmaddaford: Child answer id isn't used for relationship answers

schema/blocks/confirmation.json
```diff
@@ -0,0 +1,48 @@
...
+              "$ref": "../questions/general.json#/question"
+            },
+            {
+              "$ref": "../questions/date_range.json#/question"
+            },
+            {
+              "$ref": "../questions/repeating_answer.json#/question"
+            },
+            {
+              "$ref": "../questions/relationship.json#/question"
```
ajmaddaford: Although this is possible, we only ever use the general question type in a confirmation block, so we should restrict it here to ensure no other question types are used. 

We will create some proper structure for the confirmation block when the confirmation/summary pattern is finalised.

schema/blocks/interstitial.json
```diff
@@ -0,0 +1,96 @@
...
+        "properties": {
+          "id": {
+            "$ref": "../common_definitions.json#/id"
+          },
+          "content": {
+            "$ref": "../common_definitions.json#/guidance_content"
+          }
+        }
+      },
+      "secondary_content": {
```
ajmaddaford: primary_content, preview_content and secondary_content are specific to introduction blocks.

schema/blocks/questionnaire.json
```diff
@@ -0,0 +1,58 @@
...
+      },
+      "description": {
+        "type": "string"
+      },
+      "number": {
+        "type": "string"
+      },
+      "type": {
+        "type": "string",
+        "pattern": "^Questionnaire$"
```
ajmaddaford: I think we need to raise a card for renaming this block type from Questionnaire to Question. What do you think?

jonnyshaw89: Yeh, given we not only have a single question per page this probably makes sense

jonnyshaw89: https://trello.com/c/wbHsWRLp/430-rename-block-type-from-questionnaire-to-question

jonnyshaw89:
@ajmaddaford having looked at splitting out the 'base' schemas. i'm not sure the value it adds is more than the functionality that it takes away. For example it wont allow us to restrict on additionalProperties. Therefore not effectively allowing us to validate that a TextField isn't specifying options.

schemas/answers/checkbox.json
```diff
@@ -0,0 +1,84 @@
...
+            "value"
+          ]
+        }
+      },
+      "mandatory": {
+        "type": "boolean"
+      },
+      "alias": {
+        "type": "string",
+        "description": "An alias that can be used when piping this answer."
```
ajmaddaford: Elements that are common and have a description or restrictions e.g. child_answer_id or alias, should be defined in a common place. This will ensure we don't don't duplicate this information. We might want to think about having something like `schemas/answers/definitions.json` rather than overload the top level definitions.

PR #10, opened 2017-12-27
Rename id to block in goto routing rules

yoganandkunche:
updated Id to block in the schema and added additional properties

### DIFF
--- a/schemas/common_definitions.json
+++ b/schemas/common_definitions.json
@@ -193,7 +193,7 @@
         "goto": {
           "type": "object",
           "properties": {
-            "id": {
+            "block": {
               "type": "string",
               "description": "The id of another block in the same group."
             },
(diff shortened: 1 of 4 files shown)


schemas/common_definitions.json
```diff
@@ -191,9 +191,9 @@
         "goto": {
           "type": "object",
           "properties": {
-            "id": {
+            "block": {
               "type": "string",
-              "description": "The id of another block in the same group."
+              "description": "The block in the same group."
```
jonnyshaw89: This description isn't very good. you should describe better what this value is and how it is used.

schemas/common_definitions.json
```diff
@@ -216,8 +216,9 @@
             }
           ],
           "required": [
```
jonnyshaw89: we shouldn't need this

PR #11, opened 2018-01-02
Add view_submitted_responses to Questionnaire

dwyeradam:
Adds view_submitted_responses attribute to main questionnaire.
**enabled** is a boolean to allow viewing of submitted responses for a questionnaire.
**timeout** is an integer of the number of seconds the link for submitted answers is available after submission.

### DIFF
--- a/schemas/questionnaire_v1.json
+++ b/schemas/questionnaire_v1.json
@@ -59,6 +59,24 @@
         "StatisticsOfTradeAct"
       ]
     },
+    "view_submitted_responses": {
+      "type": "object",
+      "properties": {
+        "enabled": {
+          "description": "Whether or not answers are available after submission.",
+          "type": "boolean"
+        },
+        "timeout": {
+          "description": "The amount of time in seconds submitted answers are available for.",
+          "type": "integer"
+        }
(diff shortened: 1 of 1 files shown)


schemas/blocks/interstitial.json
```diff
@@ -42,7 +42,7 @@
         }
       },
       "content": {
-        "$ref": "common_definitions.json#/content"
+        "$ref": "../common_definitions.json#/content"
```
shahi645: @dwyeradam i think this change is in Yogi's PR

schemas/questionnaire_v1.json
```diff
@@ -59,6 +59,19 @@
...
+        "enabled": {
+          "description": "Whether or not answers are available after submission.",
+          "type": "boolean"
+        },
+        "timeout": {
+          "description": "The amount of time in seconds submitted answers are available for.",
+          "type": "integer"
+        }
+      }
+    },
```
jonnyshaw89: Are both of these properties required?
we should also add `"additionalProperties": false,` so prevent additional properties being added to the object

PR #13, opened 2018-01-10
Eq 1396 add validation for min and max value answers

shahi645:
**Context**
We wanted to add validation, so that if there are two linked answers, the range of the subsequent answer, is within the range of the used answer. 

**To check:**
- Edit `test_invalid_in_max_exclusivity` with test case that should pass/fail and the validation should run.
- Run validation against SR MWSS should throw a couple of schema validation errors as [https://github.com/ONSdigital/eq-survey-runner/pull/1405](url) is not in yet

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -39,6 +39,8 @@ class Validator:
 
         errors.extend(self.validate_child_answers_define_parent(json_to_validate))
 
+        errors.extend(self._validate_min_max_exclusivity(json_to_validate))
+
         return errors
 
     def _validate_json_against_schema(self, json_to_validate):
@@ -148,6 +150,24 @@ class Validator:
                 for used_answer_id in used_answers:
                     errors.extend(self._validate_range_type(
                         json_to_validate, used_answer_id, answer_id, answer_decimals))
+
+        return errors
+
+    def _validate_min_max_exclusivity(self, json_to_validate):
+        errors = []
+        for block in self._get_blocks(json_to_validate):
+            for answer in self._get_answers_for_block(block):
+                max_value = answer.get('max_value')
+                if max_value and max_value.get('answer_id'):
+                    if max_value.get('exclusive') and max_value.get('exclusive') is True:
+                        for block_a in self._get_blocks(json_to_validate):
+                            for answer_a in self._get_answers_for_block(block_a):
+                                if answer_a.get('id') == max_value.get('answer_id'):
+                                    if not answer_a.get('exclusive'):
+                                        error_message = '{} has exclusive set to true, {} must be exclusive also' \
+                                            .format(answer.get('id'), answer_a.get('id'))
+
+                                        errors.append(self._error_message(error_message))
 
         return errors
 
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -176,53 +176,86 @@ class Validator:
 
         if maximum:
             if max_exclusive is True:
-                if max_value:
-                    max_range = max_value - (1 / 10 ** answer_decimals)
-                elif maximum.get('answer_id'):
-                    max_value_answer_id = answers_by_id.get(maximum.get('answer_id'))
-                    range_for_answer = self._get_range_for_answer(max_value_answer_id, answers_by_id)
-                    max_range = range_for_answer[1]
+                max_range = self._get_exclusive_range(max_value, maximum, answers_by_id, answer_decimals)
+                # if max_value:
+                #     max_range = max_value - (1 / 10 ** answer_decimals)
+                # elif maximum.get('answer_id'):
+                #     max_value_answer_id = answers_by_id.get(maximum.get('answer_id'))
+                #     range_for_answer = self._get_range_for_answer(max_value_answer_id, answers_by_id)
+                #     max_range = range_for_answer[1]
 
-                elif not max_exclusive:
-                    if max_value:
-                        max_range = max_value
-                    elif maximum.get('answer_id'):
-                        max_value_answer_id = answers_by_id.get(maximum.get('answer_id'))
-                        range_for_answer = self._get_range_for_answer(max_value_answer_id, answers_by_id)
-                        max_range = range_for_answer[1]
+            elif not max_exclusive:
+                max_range = self._get_exclusive_range(max_value, maximum, answers_by_id, answer_decimals)
+                # if max_value:
+                #     max_range = max_value
+                # elif maximum.get('answer_id'):
+                #     max_value_answer_id = answers_by_id.get(maximum.get('answer_id'))
+                #     range_for_answer = self._get_range_for_answer(max_value_answer_id, answers_by_id)
+                #     max_range = range_for_answer[1]
         elif minimum:
             if min_exclusive is True:
-                if min_value:
-                    min_range = min_value + (1 / 10 ** answer_decimals)
-                elif minimum.get('answer_id'):
-                    min_value_answer_id = answers_by_id.get(minimum.get('answer_id'))
-                    range_for_answer = self._get_range_for_answer(min_value_answer_id, answers_by_id)
-                    min_range = range_for_answer[1]
+                min_range = self._get_exclusive_range(min_value, minimum, answers_by_id, answer_decimals)
+                # if min_value:
+                #     min_range = min_value + (1 / 10 ** answer_decimals)
+                # elif minimum.get('answer_id'):
+                #     min_value_answer_id = answers_by_id.get(minimum.get('answer_id'))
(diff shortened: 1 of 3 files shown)


tests/schemas/test_invalid_min_max_exclusivity.json
```diff
@@ -0,0 +1,64 @@
...
+                    ]
+                }]
+            },
+            {
+                "type": "Summary",
+                "id": "summary"
+            }
+        ],
+        "id": "group",
+        "title": ""
```
MIslam92: If this has no value being used for this you can remove this?

app/validation/validator.py
```diff
@@ -148,9 +148,105 @@ def validate_range_types_from_answers(self, json_to_validate):
...
         return errors
 
+    def _get_min_max_answer_range(self, json_to_validate, used_answers):
+        answer_ranges = {}
+        used_answer = {}
+
+        for block in self._get_blocks(json_to_validate):
+            for answer in self._get_answers_for_block(block):
+                if answer.get('id') in used_answers:
+                    used_answer[answer.get('id')] = answer
```
MIslam92: Should this not be `answer = used_answer[answer.get('id')]`

tests/schemas/test_invalid_min_max_exclusivity.json
shahi645: it is a required property for groups, but can move it under group to make more readable

MIslam92: @shahi645 Still think you should give it a relevant title as we do this currently throughout our code

app/validation/validator.py
```diff
@@ -151,6 +153,75 @@ def validate_range_types_from_answers(self, json_to_validate):
...
+        for block in self._get_blocks(json_to_validate):
+            for answer in self._get_answers_for_block(block):
+                answers[answer.get('id')] = answer
+
+        for answer_id, answer in answers.items():
+            referenced_answer_id = self._get_referenced_answer_id(answer)
+
+            if referenced_answer_id:
+                answer_min_max = self._get_range_for_answer(answer, answers)
+                # referenced_answer_range = self._get_range_for_answer(answers[referenced_answer_id], answers)
```
ajmaddaford: The commented out line should be removed

app/validation/validator.py
```diff
@@ -151,6 +153,75 @@ def validate_range_types_from_answers(self, json_to_validate):
...
+                answer_range = answer_min_max[1] - answer_min_max[0]
```
ajmaddaford: We could assign multiple variables from the `_get_range_for_answer` call:
```
minimum, maximum = self._get_range_for_answer(answer, answers)
```
so we can remove the need for using the `0` and `1` array accesses

app/validation/validator.py
```diff
@@ -151,6 +153,75 @@ def validate_range_types_from_answers(self, json_to_validate):
...
+        min_value = 0
+
+        if maximum:
+            max_value = self._get_maximum_value(answer, answers, min_value)
+        if minimum:
+            min_value = self._get_minimum_value(answer, answers, max_value)
+
+        return min_value, max_value
+
+    def _get_maximum_value(self, answer, answers, min_value):
```
ajmaddaford: I don't think the naming of this method is quite right. If `value` is set, it returns the maximum possible value for the answer. If `answer_id` is set the maximum value becomes a range (the min to max of the referenced answer), and it returns the minimum possible value from that range. 

Also, min_value doesn't appear to be used.

app/validation/validator.py
```diff
@@ -151,6 +153,75 @@ def validate_range_types_from_answers(self, json_to_validate):
...
+            value = self._get_range_for_answer(referenced_answer, answers)[0]
+
+        if 'exclusive' in maximum and maximum['exclusive'] is True:
+            max_value = value - (1 / 10 ** answer_decimals)
+        else:
+            max_value = value
+
+        return max_value
+
+    def _get_minimum_value(self, answer, answers, max_value):
```
ajmaddaford: max_value doesn't appear to be used

app/validation/validator.py
```diff
@@ -151,6 +153,75 @@ def validate_range_types_from_answers(self, json_to_validate):
...
+        answer_decimals = answer.get('decimal_places', 0)
+        maximum = answer.get('max_value')
+
+        if 'value' in maximum:
+            value = maximum.get('value')
+        elif 'answer_id' in maximum:
+            referenced_answer = answers[maximum['answer_id']]
+            value = self._get_range_for_answer(referenced_answer, answers)[0]
+
+        if 'exclusive' in maximum and maximum['exclusive'] is True:
```
ajmaddaford: This would be clearer using the dict `get` method:
```
if maximum.get('exclusive'):
```

app/validation/validator.py
```diff
@@ -151,6 +153,75 @@ def validate_range_types_from_answers(self, json_to_validate):
...
+
+    def _get_maximum_value(self, answer, answers, min_value):
+        answer_decimals = answer.get('decimal_places', 0)
+        maximum = answer.get('max_value')
+
+        if 'value' in maximum:
+            value = maximum.get('value')
+        elif 'answer_id' in maximum:
+            referenced_answer = answers[maximum['answer_id']]
+            value = self._get_range_for_answer(referenced_answer, answers)[0]
```
ajmaddaford: This would be clearer using `_` to throw away the variable you're not interested in:
```
minimum, _ = self._get_range_for_answer(referenced_answer, answers)
```

Although it would probably be better to move the default minimum value (0) into the `_get_minimum_value` method and call that directly here.

app/validation/validator.py
```diff
@@ -277,6 +348,17 @@ def _get_answers_for_block(block_json):
...
         return answers
 
+    @staticmethod
+    def _get_referenced_answer_id(answer):
+        referenced_answer_id = None
+
+        if answer.get('max_value') and 'answer_id' in answer.get('max_value'):
+            referenced_answer_id = answer['max_value']['answer_id']
+        if answer.get('min_value') and 'answer_id' in answer.get('min_value'):
+            referenced_answer_id = answer['min_value']['answer_id']
```
ajmaddaford: What should happen if both min and max values reference another answer? If we're not going to support this we should add validation to stop it.

app/validation/validator.py
```diff
@@ -151,6 +153,75 @@ def validate_range_types_from_answers(self, json_to_validate):
...
+
+        for answer_id, answer in answers.items():
+            referenced_answer_id = self._get_referenced_answer_id(answer)
+
+            if referenced_answer_id:
+                answer_min_max = self._get_range_for_answer(answer, answers)
+                # referenced_answer_range = self._get_range_for_answer(answers[referenced_answer_id], answers)
+                answer_range = answer_min_max[1] - answer_min_max[0]
+                if answer_range <= 0:
+                    error_message = 'The range of {} is outside the range of {}'.format(answer_id, referenced_answer_id)
```
ajmaddaford: It would be clearer to say "The range of {} _could_ be outside the range of {}"

app/validation/validator.py
```diff
@@ -151,6 +153,75 @@ def validate_range_types_from_answers(self, json_to_validate):
...
 
+    def _validate_min_max_ranges(self, json_to_validate):
+        errors = []
+        answers = {}
+
+        for block in self._get_blocks(json_to_validate):
+            for answer in self._get_answers_for_block(block):
+                answers[answer.get('id')] = answer
+
+        for answer_id, answer in answers.items():
```
jonnyshaw89: Do you need this 2nd loop? if you remove this and nest the code below inside the previous loop it should keep working.
As any referenced answer should be before the answer that it is referenced from?

app/validation/validator.py
jonnyshaw89: can you not just assign this to `max_value`
as it will save you the subsequent `else`

app/validation/validator.py
shahi645: i think i did have this before, but for cases where there is no it throws `none type has no 'get' attribute`

shahi645:
@dwyeradam that is what i was hoping to do at first, but the whole range stuff became far more complex for me. Also the `used_answers` in the validate range type only ever has 1 or none `used_answer_id` within it

app/validation/validator.py
shahi645: @ajmaddaford worth doing a test case to see, are we also referencing from two different answers, or the same answer?

dwyeradam:
@shahi645  Do you want to pair up on this? We can build up a structure with all information we'll need to run each of the different validations at the end.

shahi645:
@dwyeradam Yes please 😄

shahi645:
Closing due to refactor of range validation

PR #15, opened 2018-01-19
Rename view_submitted_response

dwyeradam:
Change view_submitted_response**s** to view_submitted_response
Change **timeout** to **duration**

### DIFF
--- a/schemas/questionnaire_v1.json
+++ b/schemas/questionnaire_v1.json
@@ -59,14 +59,14 @@
         "StatisticsOfTradeAct"
       ]
     },
-    "view_submitted_responses": {
+    "view_submitted_response": {
       "type": "object",
       "properties": {
         "enabled": {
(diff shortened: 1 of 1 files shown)


schemas/questionnaire_v1.json
```diff
@@ -59,14 +59,14 @@
...
-    "view_submitted_responses": {
+    "view_submitted_response": {
       "type": "object",
       "properties": {
         "enabled": {
           "description": "Whether or not answers are available after submission.",
           "type": "boolean"
         },
-        "timeout": {
+        "duration": {
```
shahi645: @dwyeradam this could be named `view_submitted_reponse_timeout`?

jonnyshaw89: There isn't much point in concatenating the names if they are nested?

PR #17, opened 2018-01-24
Range validation refactor

shahi645:
Refactor of range type validation in validator. 

Added validation to make sure possible ranges in answer cannot be zero or lower.

To test alter a schema to fail range validation and run against schema-validator

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -11,7 +11,6 @@ MAX_DECIMAL_PLACES = 6
 
 
 class Validator:
-
     def __init__(self):
         with open('schemas/questionnaire_v1.json', encoding='utf8') as schema_data:
             self.schema = load(schema_data)
@@ -29,7 +28,7 @@ class Validator:
 
         errors.extend(self.validate_routing_rules_has_default_if_not_all_answers_routed(json_to_validate))
 
-        errors.extend(self.validate_range_types_from_answers(json_to_validate))
+        errors.extend(self.validate_numeric_answer_types(json_to_validate))
 
         ignored_keys = ['routing_rules', 'skip_conditions']
         errors.extend(self.validate_duplicates(json_to_validate, ignored_keys, 'id'))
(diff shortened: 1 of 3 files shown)


shahi645:
When running SR against this branch, keep seeing this error where i think its wanting to throw the range validation error
screen shot 2018-01-26 at 11 14 36

nealedj:
I'm not seeing that syntax error in this branch. The line it refers to looks correct to me: `elif isinstance(value, dict):`

shahi645:
@MIslam92 not getting this lint error either might be something done locally

MIslam92:
Yeah disregard it then dw

app/validation/validator.py
```diff
@@ -122,80 +122,77 @@ def validate_answer(self, block, answer):
 
     def validate_range_types_from_answers(self, json_to_validate):
         errors = []
+        answer_ranges = {}
```
nealedj: Is `answer_ranges` necessary? The `answer_range` returned by `self._get_answer_ranges()` never gets used outside the scope of the loop. So maybe remove this and replace `answer_ranges[answer.get('id')]` with a single `answer_range` variable.

nealedj: I missed that fact that `_get_answer_range_values()` has `answer_ranges` passed through to it. And that seems to be to follow the `answer_id` and get the min/max values of a related answer (in `_get_answer_minimum_or_maximum()`).

Since `answer_ranges` is only getting built up within this loop there's an assumption that if `_get_answer_range_values` needs to reference another answer then that answer will already have been looped over. I.e. it assumes that answers are in order and that only previous answers will ever be referenced.

This sounds like it's taken into account but I thought I'd ask the question. @jonnyshaw89 might know?

shahi645: this is the case @nealedj, hence why we pass the ranges through

tests/test_schema_validation.py
```diff
@@ -55,6 +55,18 @@ def test_schemas(self):
             for error in errors:
                 logger.error(error)
 
+    def test_invalid_answer_range(self):
+
+        file = 'schemas/test_invalid_answer_range.json'
+
+        json_file = open(os.path.join(os.path.dirname(os.path.realpath(__file__)), file), encoding='utf8')
+        json_to_validate = load(json_file)
```
jonnyshaw89: You can probably make these 2 lines into a reusable function for the rest of the tests here?

app/validation/validator.py
```diff
@@ -122,80 +122,79 @@ def validate_answer(self, block, answer):
...
+        values = []
+        if 'value' in attributes:
+            if not MIN_NUMBER <= attributes.get('value') <= MAX_NUMBER:
+                error_message = 'Value {} used in {} should be between system limits {} to {}' \
+                    .format(attributes['value'], answer_id, MIN_NUMBER, MAX_NUMBER)
+                raise ValidationError(error_message)
+            values = [attributes.get('value')]
+
+        elif 'answer_id' in attributes:
+            if decimal_places < answer_ranges[attributes['answer_id']]['decimal_places']:
```
jonnyshaw89: what if the `answer_id` isn't in `answer_ranges`

app/validation/validator.py
```diff
@@ -122,80 +122,79 @@ def validate_answer(self, block, answer):
...
+                                                       'minimum'),
+            'max': self._get_answer_minimum_or_maximum(answer['id'],
+                                                       answer_ranges,
+                                                       answer.get('max_value', {'value': MAX_NUMBER}),
+                                                       answer.get('decimal_places', 0),
+                                                       'maximum'),
+            'decimal_places': answer.get('decimal_places', 0),
+        }
+
+    def _get_answer_minimum_or_maximum(self, answer_id, answer_ranges, attributes, decimal_places, min_or_max):
```
jonnyshaw89: `attributes` here is not very clear as to what it actually is

app/validation/validator.py
```diff
@@ -122,80 +122,79 @@ def validate_answer(self, block, answer):
...
+                error_message = '{} has {} decimal places and is being passed to max/min value for {} of {} ' \
+                                'decimal_places. The referred answer cannot have more decimal places'\
+                    .format(attributes['answer_id'], answer_ranges[attributes['answer_id']]['decimal_places'],
+                            answer_id, decimal_places)
+                raise ValidationError(error_message)
+            values = self._get_range_for_referred_answer(answer_ranges, attributes['answer_id'])
+
+        for idx, value in enumerate(values):
+            if isinstance(value, list):
+                values[idx] = value[0]
```
jonnyshaw89: I don't really understand what these 3 lines are doing?

app/validation/validator.py
shahi645: if the answer id is not an existing id of an answer, the method `_get_rsnge_for_referred_answer` will raise an erro

jonnyshaw89: which is called after this?

app/validation/validator.py
shahi645: So when there is an answer_id the method goes and gets that answers [[min], [max]] and they form the new minimum or maximum as possibilities, but if its just a value its returned as [min] or [max].
So these lines de-list and then put into a new list, which means that math operations can occur when they are exclusive or not

app/validation/validator.py
shahi645: It is basically the attribute of whether the answer has a min_value or a max_value

app/validation/validator.py
jonnyshaw89: but if the value is a list it only uses the 1st value? therefore why create a list in the first place if only the 1st value will be used.

app/validation/validator.py
jonnyshaw89: but the name `attributes` is too vague. i had to look at the calling method to have any idea what it was.

shahi645: how about `min_max_attribute`?

app/validation/validator.py
shahi645: Yes but this checks within a `min_value` or a `max_value`, so in that case it will always be referring a previous answer_id

app/validation/validator.py
jonnyshaw89: Im not really a fan of the whole min/max in the names of thing. you are naming the function based on the values you are passing in and not on what it actually does

app/validation/validator.py
jonnyshaw89: your assuming its always a previous answer. if the schema was incorrect ( which is likely hence we are validating it) then we would want to return an error saying that you are referencing  future answer rather then blowing up with an index error

app/validation/validator.py
shahi645: So the attributes being passed in can vary from values, to a dict, so maybe something like `range_attributes`

jonnyshaw89: if what gets passed in car vary in type then that is also bad

app/validation/validator.py
shahi645: Right so it should go off to validate that answer_id exists before the decimal places validation

jonnyshaw89: would it work if you moved `_get_range_for_referred_answer` before this?

dwyeradam:
We're at a point where range validation is doing five different checks within three different methods. This can make it quite difficult to understand what is actually being tested by `validate_range_types_from_answers`. Is it worth considering looking at creating the answer_ranges object and then doing as much validation against that at the same point. We currently have checks on lines 140, 146, 172, 179 and 289.

app/validation/validator.py
```diff
@@ -120,82 +119,45 @@ def validate_answer(self, block, answer):
...
-                if answer.get('max_value') and 'answer_id' in answer.get('max_value'):
-                    used_answers.append(answer['max_value']['answer_id'])
+            # Validate referred numeric answer exists (skip further tests for answer if error is returned)
+            if self._validate_referred_numeric_answer(answer, answer_ranges, errors):
+                continue
 
-                if answer.get('min_value') and 'value' in answer.get('min_value'):
-                    values.append(answer['min_value']['value'])
+            # Validate numeric answer has a positive range of possible responses
+            self._validate_numeric_range(answer, answer_ranges, errors)
```
jonnyshaw89: We should return a list of errors from this function and extend the errors list.
Modifying a function parameter is bad practice and its not clear that that parameter will be updated

app/validation/validator.py
```diff
@@ -277,6 +246,111 @@ def _get_answers_for_block(block_json):
...
+        if defined_minimum and defined_minimum.get('exclusive') and minimum:
+            exclusive_minimum = []
+            for value in minimum:
+                exclusive_minimum.append(value + (1 / 10 ** answer.get('decimal_places', 0)))
+            minimum = exclusive_minimum
+
+        return minimum
+
+    @staticmethod
+    def _get_answer_maximum(answer, answer_ranges):
```
jonnyshaw89: there is a lot of common code between this and `_get_answer_minimum` any chance of calling a 3rd function to do the common logic?

dwyeradam: @jonnyshaw89 I've split it out so see what you think. To simplify it i've done it in two steps. One that gets the defined value, then another to convert that to exclusive if required.

dwyeradam:
@MIslam92  ready for new review.

app/validation/validator.py
```diff
@@ -277,6 +248,122 @@ def _get_answers_for_block(block_json):
...
+            errors.append(self._error_message(error_message))
+
+        return errors
+
+    def _validate_numeric_range(self, answer, answer_ranges):
+        errors = []
+        for max_value in answer_ranges[answer.get('id')]['max']:
+            for min_value in answer_ranges[answer.get('id')]['min']:
+                if max_value - min_value < 0:
+                    error_message = 'Invalid range of min = {} and max = {} is possible for answer "{}".'\
```
MIslam92: Would this not read better as `Invalid range: The answer "{}" must be between the min = {} and max = {}`

MIslam92:
@dwyeradam "Squishy Squashy"

PR #18, opened 2018-01-31
New validator for dropdown input type

MebinAbraham:
The drop-down input type requires a new schema validator to ensure that the schemas are accurate.

Card: https://trello.com/c/lZqYhhjw/1328-add-dropdown-input-type-m

### How to review

1. In eq-survey-runner, checkout to branch: `eq-1328-add-dropdown-input-type`
2. Run the following:
`./scripts/test_schemas.sh data/en/`

Check that all the schemas pass

### DIFF
--- a/schemas/answers/dropdown.json
+++ b/schemas/answers/dropdown.json
@@ -1,0 +1,56 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "answer": {
+    "type": "object",
+    "properties": {
+      "id": {
+        "$ref": "../common_definitions.json#/id"
+      },
+      "q_code": {
+        "$ref": "../common_definitions.json#/q_code"
+      },
+      "guidance": {
+        "$ref": "../common_definitions.json#/guidance"
+      },
+      "type": {
+        "type": "string",
+        "enum": [
+          "Dropdown"
+        ]
+      },
+      "mandatory": {
+        "type": "boolean"
+      },
+      "label": {
+        "type": "string"
+      },
+      "description": {
+        "type": "string"
+      },
+      "options": {
+        "$ref": "definitions.json#/options"
+      },
+      "alias": {
+        "$ref": "definitions.json#/alias"
+      },
+      "validation": {
+        "type": "object",
+        "properties": {
+          "messages": {
+            "MANDATORY_DROPDOWN": {
+              "type": "string"
+            }
+          }
+        }
+      }
+    },
+    "additionalProperties": false,
+    "required": [
+      "id",
+      "type",
+      "mandatory",
+      "label",
+      "options"
+    ]
+  }
+}
--- a/schemas/common_definitions.json
+++ b/schemas/common_definitions.json
@@ -134,6 +134,9 @@
     "type": "object",
     "description": "These messages override the standard error messages.",
     "properties": {
+      "MANDATORY_DROPDOWN": {
+        "type": "string"
+      },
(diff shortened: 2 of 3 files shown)


shahi645:
@MebinAbraham This is good, but we should add validation about character limits within each drop down

jonnyshaw89:
@shahi645 is that something we want to restrict here?

MebinAbraham:
@shahi645 I didn't add any validation for that currently as it is not clear whether it is necessary by the business. I believe this is something that will be discussed at a later stage. Once it has been decided, it can be added if necessary.

schemas/answers/dropdown.json
```diff
@@ -0,0 +1,58 @@
...
+      "mandatory": {
+        "type": "boolean"
+      },
+      "label": {
+        "type": "string"
+      },
+      "description": {
+        "type": "string"
+      },
+      "default_value": {
```
jonnyshaw89: Do we use this value in runner?

MebinAbraham: My naming wasn't probably great but default_value is referring to the default drop-down option eg "Please select an answer".

shahi645:
@jonnyshaw89 It will be going forward. if it is not a requirement we should leave it for another time

schemas/answers/dropdown.json
```diff
@@ -0,0 +1,58 @@
...
+      "mandatory": {
+        "type": "boolean"
+      },
+      "label": {
+        "type": "string"
+      },
+      "description": {
+        "type": "string"
+      },
+      "preselected_option": {
```
jonnyshaw89: I don't really like how this is used to inject a value into the options list.
We shouldn't ever want to pre-select a value for someone

@ajmaddaford what are you views on this?

nealedj: The SDC spec says that the initially-selected option should always be the null "Select an option".

I guess this might be necessary later for pre-population though?

shahi645: I think the AC on the trello card wants this value to be configurable:
https://trello.com/c/lZqYhhjw/1328-add-dropdown-input-type-m

jonnyshaw89: @nealedj if its not needed now then we shouldn't have it in the code.
@shahi645 Maybe it could be the Label?

nealedj: @jonnyshaw89 Ah I misunderstood what this actually does. So this is just where the default option text gets set? Surely something like `default_text` makes it pretty clear?

Label could be confused with the form input label.

shahi645: @MebinAbraham has asked Charlotte whether it is required.

MebinAbraham: I've followed this up with Charlotte, and she said that she will update the ticket as we do not need to override the text. It should just always just say "Select an answer".
Ill sort this shortly.

MIslam92:
@jonnyshaw89 Can you have a re-review of this? As your review is the last blocking review

schemas/answers/dropdown.json
```diff
@@ -0,0 +1,57 @@
...
+        }
+      }
+    },
+    "additionalProperties": false,
+    "required": [
+      "id",
+      "type",
+      "mandatory",
+      "label",
+      "description",
```
ajmaddaford: `description` should be optional

nealedj:
@MebinAbraham All looks good - okay to squash please?

PR #26, opened 2018-02-20
Replace pylint-mccabe with pylint built-in complexity checker

jonnyshaw89:
We use pylint-mccabe to do complexity checking on our Python code. Recently this has become incompatible with the latest version of astroid (used by pylint) - infoxchange/pylint-mccabe#3. It also hasn't been updated in over 5 years.

We should replace it with the pylint built in complexity checker - https://pylint.readthedocs.io/en/latest/technical_reference/extensions.html#design-checker (as mentioned in PyCQA/astroid#471)

### How to review
Does the linting work?

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -20,7 +20,7 @@ persistent=yes
 
 # List of plugins (as comma separated values of python modules names) to load,
 # usually to register additional checkers.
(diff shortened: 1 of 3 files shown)


shahi645:
@jonnyshaw89 squishy squashy

PR #29, opened 2018-03-05
date_comparison routing definition

dwyeradam:
Addition of `date_comparison` to common `when` routing definition.

### DIFF
--- a/schemas/common_definitions.json
+++ b/schemas/common_definitions.json
@@ -110,6 +110,61 @@
             "boolean",
             "integer"
           ]
+        },
+        "date_comparison": {
+          "type": "object",
+          "properties": {
+            "id": {
+              "type": "string"
+            },
+            "meta": {
+              "type": "string",
+              "description": "Metadata provided by the calling service. This will vary between surveys."
+            },
+            "value": {
+              "type": "string",
+              "description": "Hard coded YYYY-MM-DD/YYYY-MM date or now to give current date",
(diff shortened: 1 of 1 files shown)


schemas/common_definitions.json
```diff
@@ -110,6 +110,49 @@
...
+              "properties": {
+                "days": {
+                  "type": "integer"
+                },
+                "months": {
+                  "type": "integer"
+                },
+                "years": {
+                  "type": "integer"
+                }
```
jonnyshaw89: You should require at lease one of these 3

schemas/common_definitions.json
```diff
@@ -110,6 +110,49 @@
...
+            "id": {
+              "type": "string"
+            },
+            "meta": {
+              "type": "string",
+              "description": "Metadata provided by the calling service. This will vary between surveys."
+            },
+            "value": {
+              "type": "string",
+              "description": "Hard coded YYYY-MM-DD/YYYY-MM date or now to give current date"
```
jonnyshaw89: You can add a regex to this

schemas/common_definitions.json
```diff
@@ -110,6 +110,61 @@
             "boolean",
             "integer"
           ]
+        },
+        "date_comparison": {
+          "type": "object",
+          "properties": {
+            "id": {
+              "type": "string"
```
dwyeradam: I'll need to add a validation check that only id's of a date type can be used here.

shahi645: @dwyeradam Happy with this, but will wait for your validator changes before i approve

dwyeradam: This will now be done in a separate PR that refactors all validation around routing rules.

PR #30, opened 2018-03-07
Eq 1868 validate sum against total

shahi645:
Added new question type schema `Calculated`. 

Test against https://github.com/ONSdigital/eq-survey-runner/pull/1485

### DIFF
--- a/schemas/questions/calculated.json
+++ b/schemas/questions/calculated.json
@@ -1,0 +1,100 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "question": {
+    "type": "object",
+    "properties": {
+      "id": {
+        "$ref": "../common_definitions.json#/id"
+      },
+      "title": {
+        "type": "string"
+      },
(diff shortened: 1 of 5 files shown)


schemas/questions/calculated.json
```diff
@@ -0,0 +1,100 @@
...
+            "enum": [
+              "sum",
+              "average"
+            ]
+          },
+          "answer_id": {
+            "type": "string",
+            "description": "The id of an answer from which to obtain the total to validate against"
+          },
+          "answers_to_group": {
```
MebinAbraham: Need to check that min no. of answers in the answer block also matches this.

schemas/questions/calculated.json
```diff
@@ -0,0 +1,100 @@
...
+            "type": "array",
+            "minItems": 2,
```
MebinAbraham: Need to check that min no. of answers in the answer block also matches this.

PR #32, opened 2018-03-08
Improve validation of answer options

MebinAbraham:
Enforces that any answer type that uses options, ie Radios, Checkboxes, Dropdowns etc must contain at least 2 options. Also checks for duplicate option label and value between the options.

### How to review 
1. Ensure schema validation fails when an answer type that uses options doesn't have at least 2 options and vice versa.
2. Ensure that validation fails when options contain duplicates and vice versa.

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -36,6 +36,8 @@ class Validator:
         errors.extend(self.validate_duplicates(json_to_validate, ignored_keys, 'id'))
         errors.extend(self.validate_duplicates(json_to_validate, ignored_keys, 'alias'))
 
+        errors.extend(self.validate_duplicate_options(json_to_validate))
+
         errors.extend(self.validate_contains_confirmation_or_summary(json_to_validate))
 
         errors.extend(self.validate_child_answers_define_parent(json_to_validate))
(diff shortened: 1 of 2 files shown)


app/validation/validator.py
```diff
@@ -173,6 +175,29 @@ def validate_duplicates(self, json_to_validate, ignored_keys, special_key):
...
+            for answers in answers_for_block:
+                labels = []
+                values = []
+
+                if 'options' in answers:
+                    for option in answers.get('options'):
+
+                        if option['label'] in labels:
+                            errors.append(self._error_message('Duplicate label found - {}'.format(option['label'])))
+                        elif option['value'] in values:
```
nealedj: I don't think this needs to be an `elif`? It could just be an `if` since both the label and value might be be invalid.

app/validation/validator.py
```diff
@@ -173,6 +175,29 @@ def validate_duplicates(self, json_to_validate, ignored_keys, special_key):
...
+
+        for block in self._get_blocks(json_to_validate):
+            answers_for_block = self._get_answers_for_block(block)
+
+            for answers in answers_for_block:
+                labels = []
+                values = []
+
+                if 'options' in answers:
+                    for option in answers.get('options'):
```
nealedj: Consider using `for option in answers.get('options', []):` and removing the preceding if statement. Too much nesting makes functions less readable.

app/validation/validator.py
```diff
@@ -193,6 +195,29 @@ def validate_duplicates(self, json_to_validate, ignored_keys, special_key):
...
         return duplicate_errors
 
+    def validate_duplicate_options(self, json_to_validate):
+        errors = []
+
+        for block in self._get_blocks(json_to_validate):
+            answers_for_block = self._get_answers_for_block(block)
+
+            for answers in answers_for_block:
+                labels = set()
```
jonnyshaw89: any reason this cant just be `labels = []`

app/validation/validator.py
```diff
@@ -193,6 +195,29 @@ def validate_duplicates(self, json_to_validate, ignored_keys, special_key):
...
+                values = set()
+
+                for option in answers.get('options', []):
+
+                    if option['label'] in labels:
+                        errors.append(self._error_message('Duplicate label found - {}'.format(option['label'])))
+
+                    if option['value'] in values:
```
jonnyshaw89: What if we wanted 2 different option labels that have the same value in the background?
I think we have had this before in one of the surveys when it was split on the display to make the options more refined but downstream the value was the same.

MebinAbraham: Noting for others. Having duplicate values will cause the functionality of the answer type, ie Radio buttons to not work properly. Two options with the same value mean the top option will always be chosen even if they chose the bottom option which has a different label but same value.

app/validation/validator.py
MebinAbraham: @jonnyshaw89 Since I am adding after each option, no need to add the duplicates. Would work either way but any reason not to have it as `set()` as `sets` are faster for searching I believe?

PR #33, opened 2018-03-14
change list name from answers_to_group to answers_to_calculate

shahi645:
Just a quick PR to change `answers_to_group` name to `answers_to_calculate`

### DIFF
--- a/schemas/questions/calculated.json
+++ b/schemas/questions/calculated.json
@@ -28,7 +28,9 @@
       },
       "type": {
         "type": "string",
-        "enum": ["Calculated"]
+        "enum": [
+          "Calculated"
+        ]
       },
       "calculated": {
         "type": "object",
@@ -36,15 +38,14 @@
           "calculated_type": {
             "type": "string",
             "enum": [
-              "sum",
-              "average"
+              "sum"
             ]
           },
           "answer_id": {
             "type": "string",
             "description": "The id of an answer from which to obtain the total to validate against"
           },
-          "answers_to_group": {
+          "answers_to_calculate": {
             "type": "array",
             "minItems": 2,
             "items": {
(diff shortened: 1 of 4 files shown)


schemas/questions/calculated.json
```diff
@@ -64,7 +64,7 @@
         "required": [
           "calculated_type",
           "answer_id",
-          "answers_to_group",
+          "answers_to_calculate",
```
jonnyshaw89: I think the conditions values above should match our existing condition names
https://github.com/ONSdigital/eq-schema-validator/blob/master/schemas/common_definitions.json#L96

shahi645: @jonnyshaw89 do u mean use the common definitions `conditions` or just the fact that they are a bit more readable i.e. `equal to`, `less than`, `less than or equal to`

jonnyshaw89: just use the same values. it doesn't look good where we have multiple ways of writing the same thing

shahi645: but they aren't actually the same thing, they happen to be named similarly, I actually prefer it with the hy[hens in there @nealedj second opinion ?

jonnyshaw89: so `equals` isn't the same as `equal-to`?

dwyeradam: I'm agreeing with Jon here. If **"equal-to"** is checking equality between two things then we have used **"equals"** across the schema. If **"equal-to"** is not checking the equality between two things then the naming is even worse.

dwyeradam: Also we don't have a concept of **"less-than-equal-to"**. We would specify **"less than"** or **"equals"**.  Allowing multiple conditions to be specified so we have a greater range of possibilities but with less code.

schemas/questions/calculated.json
```diff
@@ -28,73 +28,83 @@
...
-            },
-            {
-              "$ref": "../answers/number.json#/answer"
-            },
-            {
-              "$ref": "../answers/percentage.json#/answer"
+          "conditions": {
+            "type": "array",
+            "minItems": 1,
+            "maxItems": 2,
```
jonnyshaw89: Would it not be better to enforce unique than define the max?
`"uniqueItems": true,`

shahi645: @jonnyshaw89 would we want both ?  otherwise with `uniqueItems` the user could enter 3 ?

jonnyshaw89: yeh that is the case now you have added `greater than` before there were only 2 so you didn't need the max

jonnyshaw89:
@shahi645 if you have a list in your commit message it should have a title on the 1st line

PR #34, opened 2018-03-15
Add default for numeric answers

dwyeradam:
The "default" value will be stored in the answer store on POST when no value is given for the answer.


dwyeradam:
@jonnyshaw89 Guess i'll add it. Strictly speaking the test isn't needed as default only works if no answer is given which isn't possible with mandatory answers. However the test would stop people possibly expecting behaviour that wouldn't work.

PR #36, opened 2018-03-20
Refactored validator to reduce the amount of looping

jonnyshaw89:
Refactored the amount of looping and repetition within each stop of validation.

This should make it easier to add new tests going forward.


app/validation/validator.py
```diff
@@ -58,58 +73,40 @@ def _validate_json_against_schema(self, json_to_validate):
...
-        blocks = self._get_blocks(json_to_validate)
-        for block in blocks:
-            if 'routing_rules' in block and block['routing_rules']:
-                for rule in block['routing_rules']:
-                    if 'goto' in rule and 'id' in rule['goto'].keys():
-                        block_id = rule['goto']['id']
-                        if block_id == 'summary':
-                            continue
+        if 'goto' in rule and 'id' in rule['goto'].keys():
+            block_id = rule['goto']['id']
```
dwyeradam: 'id' is never in a goto rule. It should be 'block'. See: [goto definition](https://github.com/ONSdigital/eq-schema-validator/blob/master/schemas/common_definitions.json#L252)

app/validation/validator.py
```diff
@@ -24,23 +23,39 @@ def validate_schema(self, json_to_validate):
...
+                if section == json_to_validate['sections'][-1]:
+                    if group == section['groups'][-1]:
+                        errors.extend(self.validate_contains_confirmation_or_summary(group['blocks']))
 
-        errors.extend(self.validate_numeric_answer_types(json_to_validate))
+                for block in group['blocks']:
+
+                    if 'routing_rules' in block and block['routing_rules']:
+                        for rule in block['routing_rules']:
+                            errors.extend(self.validate_schema_contains_valid_routing_rules(section['groups'], group['blocks'], rule))
```
dwyeradam: I know it never has but shouldn't we be validating routing rules at a group level as well?

dwyeradam: Just to further complicate this. Should this be extended for skip conditions?

app/validation/validator.py
```diff
@@ -24,23 +23,39 @@ def validate_schema(self, json_to_validate):
...
-        ignored_keys = ['routing_rules', 'skip_conditions']
-        errors.extend(self.validate_duplicates(json_to_validate, ignored_keys, 'id'))
-        errors.extend(self.validate_duplicates(json_to_validate, ignored_keys, 'alias'))
 
-        errors.extend(self.validate_duplicate_options(json_to_validate))
+                        errors.extend(self.validate_child_answers_define_parent(question['answers']))
 
-        errors.extend(self.validate_contains_confirmation_or_summary(json_to_validate))
+                        for answer in question['answers']:
+                            errors.extend(self.validate_answer(block, answer))
```
dwyeradam: Can we name this to describe what the validation check is actually checking.

app/validation/validator.py
jonnyshaw89: added group routing

jonnyshaw89: skip conditions use when not goto

app/validation/validator.py
```diff
@@ -24,23 +23,38 @@ def validate_schema(self, json_to_validate):
...
-        ignored_keys = ['routing_rules', 'skip_conditions']
-        errors.extend(self.validate_duplicates(json_to_validate, ignored_keys, 'id'))
-        errors.extend(self.validate_duplicates(json_to_validate, ignored_keys, 'alias'))
+                    for rule in block.get('routing_rules', []):
+                        errors.extend(self.validate_schema_contains_valid_routing_rules(group['blocks'], 'block', rule))
+                        errors.extend(self.validate_schema_contains_valid_routing_rules(section['groups'], 'group', rule))
+
+                    for question in block.get('questions', []):
+
+                        errors.extend(self.validate_calculated_ids_in_answers_to_calculate_exists(question))
```
shahi645: Small change, and I'm not sure if this is a better way of doing this but what if we check `question['type']` here, instead of in the method for validating. Just a thought

app/validation/validator.py
```diff
@@ -24,23 +23,38 @@ def validate_schema(self, json_to_validate):
...
+                        errors.extend(self.validate_child_answers_define_parent(question['answers']))
```
shahi645: this can go under the next `for` loop or answer

app/validation/validator.py
```diff
@@ -24,23 +23,38 @@ def validate_schema(self, json_to_validate):
...
 
-        errors.extend(self.validate_duplicate_options(json_to_validate))
+                            if answer['type'] in ['Number', 'Currency', 'Percentage']:
+                                numeric_answer_ranges[answer.get('id')] = self._get_numeric_range_values(answer, numeric_answer_ranges)
 
-        errors.extend(self.validate_contains_confirmation_or_summary(json_to_validate))
+                                errors.extend(self.validate_numeric_answer_types(answer, numeric_answer_ranges))
 
-        errors.extend(self.validate_child_answers_define_parent(json_to_validate))
+
```
shahi645: one less blank line

app/validation/validator.py
```diff
@@ -193,123 +172,86 @@ def _get_numeric_range_values(self, answer, answer_ranges):
...
+            unique_items = []
+            duplicate_errors = []
+
+            for value in self._parse_values(json_to_validate, special_key):
+                if value in unique_items:
+                    duplicate_errors.append(self._error_message('Duplicate {} found. value {}'.format(special_key, value)))
+                else:
+                    unique_items.append(value)
 
         return duplicate_errors
```
shahi645: Should be tabbed in once

app/validation/validator.py
```diff
@@ -193,123 +172,86 @@ def _get_numeric_range_values(self, answer, answer_ranges):
...
-                            errors.extend([self._error_message('Child answer %s defines incorrect parent_answer_id %s '
-                                                               'in schemas: Should be %s'
-                                                               % (child_answer_id,
-                                                                  answers_by_id[child_answer_id]['parent_answer_id'],
-                                                                  answer_id))])
-                            continue
+        answers_by_id = self._get_answers_by_id_for_block(answers)
+
+        for answer in answers:
+            if answer['type'] in ['Radio', 'Checkbox']:
```
shahi645: Again not sure if these would be better served being above before coming into this method

shahi645:
Also getting a lint error:
`./app/validation/validator.py:17:5: C901 'Validator.validate_schema' is too complex (12)`

app/validation/validator.py
jonnyshaw89: I disagree, it spreads the logic for the specific check across 2 functions then.

app/validation/validator.py
jonnyshaw89: No it can't as it needs a list of answers

app/validation/validator.py
jonnyshaw89: No it shouldn't. otherwise it will never evaluate the 2nd iteration of the loop.
`duplicate_errors = []` should have been outside the `for`

app/validation/validator.py
jonnyshaw89: I think its better to keep it contained in the function

app/validation/validator.py
shahi645: ahh yeah! my bad

app/validation/validator.py
shahi645: ok can go with that just a thought

shahi645:
@jonnyshaw89 Doesnt validate routing to a different group in a different section. Need to allow this or we get errors

shahi645:
@jonnyshaw89 still getting a routing error for MWSS:
`HTTP Status @ /validate: [200]
Error: [{  "errors": [    {      "message": "Schema Integrity Error. Routing rule routes to invalid group [fortnightly-pay]"    },     {      "message": "Schema Integrity Error. Routing rule routes to invalid block [fortnightly-pay-significant-changes-gross-pay]"    },     {      "message": "Schema Integrity Error. Routing rule routes to invalid block [Four Weekly Pay]"    },     {      "message": "Schema Integrity Error. Routing rule routes to invalid group [five-weekly-pay]"    },     {      "message": "Schema Integrity Error. Routing rule routes to invalid block [five-weekly-pay-significant-changes-gross-pay]"    }  ]}]`

shahi645:
needs access to multiple sections also

app/validation/validator.py
```diff
@@ -26,7 +26,9 @@ def validate_schema(self, json_to_validate):  # noqa: C901  pylint: disable=too-
 
         numeric_answer_ranges = {}
 
-        all_groups = itertools.chain.from_iterable(section.get('groups', []) for section in json_to_validate['sections'])
+        all_groups = []
+        for section in json_to_validate['sections']:
+            all_groups.extend(section.get('groups'))
```
shahi645: Just put this in the below `for` removes duplication

jonnyshaw89: Can't do that as it will only have a list of previous groups then, which isn't sufficient for routing

PR #37, opened 2018-03-20
Added support for month year date range

MebinAbraham:
Added `MonthYearDate` as one of the answer types for `DateRange` question type.
Needed to support changes in runner to support `MonthYearDate` ranges.
[Branch](https://github.com/ONSdigital/eq-survey-runner/tree/eq-1931-pipe-difference-between-two-dates)


jonnyshaw89:
@MebinAbraham Squash

PR #39, opened 2018-03-22
Don't validate range for defaulted numeric

jonnyshaw89:
This gets us around the issue where a defaulted value routes around a breakdown


app/validation/validator.py
```diff
@@ -282,6 +283,8 @@ def get_defined_numeric_value(defined_value, system_default, answer_ranges):
             referred_answer = answer_ranges.get(defined_value['answer_id'])
             if referred_answer is None:
                 values = None  # Referred answer is not  valid (picked up by _validate_referred_numeric_answer)
+            elif referred_answer.get('default') is not None:
```
shahi645: @jonnyshaw89 can this just be elif referred_answer.get('default')

jonnyshaw89: No because otherwise `0` would evaluate

PR #40, opened 2018-03-23
Added confirmation question block

jonnyshaw89:


schemas/blocks/confirmation_question.json
```diff
@@ -0,0 +1,62 @@
...
+      "type": {
+        "type": "string",
+        "enum": ["ConfirmationQuestion"]
+      },
+      "questions": {
+        "type": "array",
+        "items": {
+          "oneOf": [
+            {
+              "$ref": "../questions/general.json#/question"
```
dwyeradam: Should this be only one allowed?

PR #42, opened 2018-03-27
Don't return 200 status_code on error

jonnyshaw89:
Returning a `400` error makes it easier to detect if a schema is invalid


WickyNilliams:
Needs a corresponding PR for publisher

jonnyshaw89:
@WickyNilliams So this will break publisher?

PR #43, opened 2018-04-06
Remove answer alias from schema

nealedj:
Answer alias is removed from runner in ONSdigital/eq-survey-runner#1533

Do not merge until runner PR is merged.


nealedj:
@MebinAbraham good call - have removed. My editor was adding an empty line to every file.

PR #44, opened 2018-04-09
Eq 1947 date range validation

shahi645:
Added object types `minimum/maximum` to: 

- `DateRange` question for period validation
- `Date` answer type. 

Also added further validation in answer types where there is a minimum and a maximum so that, schema shouldn't fail. Only validates where there is a minimum and a maximum and both have `value` offset dates.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -18,6 +18,7 @@ pytest = "==2.8.2"
 "flake8" = "==3.4.1"
 pylint = "==1.8.2"
 pylint-quotes = "==0.1.6"
+python-dateutil = "*"
 
 
 [requires]
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -1,7 +1,10 @@
+import re
 from json import load
 import os
 import pathlib
 
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
 from jsonschema import SchemaError, RefResolver, validate, ValidationError
 
 MAX_NUMBER = 9999999999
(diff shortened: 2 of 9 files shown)


Pipfile
```diff
@@ -18,6 +18,7 @@ pytest = "==2.8.2"
 "flake8" = "==3.4.1"
 pylint = "==1.8.2"
 pylint-quotes = "==0.1.6"
+python-dateutil = "*"
```
MIslam92: Should you specify this to a version like the rest of the dev packages?

shahi645: I don't think we need to for `dateutil`, as we haven't in SR. Will double check with David

MIslam92: Scrap that then, if its using the latest version in SR, I'm fine with that being the same on Validator.

dwyeradam: We should always try and use the latest. We would only ever pin the version if a newer version caused issues with other dependencies.

app/validation/validator.py
```diff
@@ -131,6 +143,32 @@ def validate_routing_on_answer_options(self, block, answer):
...
+            minimum_period = example_date + self.get_relative_delta_offset_date(period_limits['minimum'])
+            maximum_period = example_date + self.get_relative_delta_offset_date(period_limits['maximum'])
+
+            if minimum_period > maximum_period:
+                errors.append(self._error_message('The minimum period is greater than the maximum period for {}'.
+                                                  format(question_id)))
+
+        return errors
+
+    def validate_minimum_and_maximum_offset_date(self, answer):
```
AndrewTorrance: DocString ?

app/validation/validator.py
```diff
@@ -394,3 +432,30 @@ def _parse_values(self, schema_json, parsed_key):
...
                     if isinstance(schema_item, dict):
                         yield from self._parse_values(schema_item, parsed_key)
+
+    def _get_offset_date_value(self, answer_min_or_max):
+
+        if answer_min_or_max['value'] == 'now':
+            value = datetime.utcnow().strftime('%Y-%m-%d')
+        else:
+            value = answer_min_or_max['value']
+
```
MebinAbraham: Just me being picking and not something you havw to do:`value = datetime.utcnow().strftime('%Y-%m-%d') if answer_min_or_max['value'] == 'now' else answer_min_or_max['value']`

shahi645: I see what your saying as it does mean a tiny bit less code, in this instance i think it's better to leave as is for readability

Pipfile
nealedj: I agree with @dwyeradam - but only because we have dependency locking (piplock).

If we didn't, then I'd always favour pinning to an absolute version. Have had too many headaches in the past due to incompatible upgrades happening at times I _really_ didn't want them to 😄

app/validation/validator.py
nealedj: In this case I think it's better as it is.

Ternary expressions in python feel clunky to me (compared to the C-style `conditional ? true_value : false_value`).

As a rule of thumb I think they're best used when they fit onto one 80 char line.

Personal preference though.

PR #46, opened 2018-04-18
Pin pip and pipenv to working versions in concourse

MebinAbraham:
Concourse is failing since latest update to pip.

I have pinned the version of `pip` and `pipenv` until this issue is resolved.
Note: Concourse is still throwing the same error as before but is able to carry on.

[Issue](https://github.com/wemake-services/wemake-django-template/issues/287)

I have also separated the scripts that are run in `concourse.yml` to `install.sh` so that it could be modified without resetting the `concourse.yml` each time.

### DIFF
--- a/ci/concourse.yml
+++ b/ci/concourse.yml
@@ -57,12 +57,7 @@ jobs:
         - -exc
         - |
           cd eq-schema-validator
-          pyenv install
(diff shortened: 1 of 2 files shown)


shahi645:
@MebinAbraham Are these requirements for the smoke tests?

PR #47, opened 2018-04-19
Specify required metadata in schema

MebinAbraham:
Runner [PR](https://github.com/ONSdigital/eq-survey-runner/pull/1545) must be merged first. Refer to runner PR for more context.
This PR enforces that:

### How to review
2. Edit a schema to ensure that if you do not define the piped metadata as required then it fails.

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -78,22 +78,15 @@ class Validator:
             return '{}'.format(e)
 
     def _validate_schema_contain_required_metadata(self, schema):
-        # Ensure that metadata used within the schema are implicitly defined in required_metadata field
+        # Ensure that metadata used within the schema are explicitly defined.
 
         errors = []
-        default_metadata = ['user_id', 'period_id']
 
-        required_metadata = schema['required_metadata']
+        default_metadata = ['user_id', 'period_id']  # Metadata that must be present in all schemas
+        required_metadata = schema['required_metadata']  # Metadata required within the schema.
 
-        # This regex will find all words that precede with "metadata" then a "[" or "." following by an optional "'"
+        # Find all words that precede with "metadata['"
         parsed_metadata = re.findall(r"(?<=metadata\[\')\w+", str(schema))
-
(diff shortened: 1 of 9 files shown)


app/validation/validator.py
```diff
@@ -74,6 +77,27 @@ def _validate_json_against_schema(self, json_to_validate):
...
+        required_metadata = schema['required_metadata']  # Metadata required within the schema.
+
+        # Find all words that precede with "metadata['"
+        parsed_metadata = re.findall(r"(?<=metadata\[\')\w+", str(schema))
+
+        for metadata in set(parsed_metadata):
+            if metadata not in required_metadata:
+                errors.append(self._error_message('Metadata - {} not specified in required_metadata field'.format(metadata)))
+
+        for metadata in required_metadata:
```
nealedj: I don't think this check adds any value and could cause author issues in the future.

If somebody were to add some required metadata I'm not sure it makes sense to force them to make use of it.

schemas/questionnaire_v1.json
```diff
@@ -59,6 +60,10 @@
         "StatisticsOfTradeAct"
       ]
     },
+    "required_metadata": {
```
nealedj: See comment on runner PR about renaming this to `metadata`

app/validation/validator.py
MebinAbraham: I get your point but I thought the point was so that all metadata needed to launch and submit the survey must be specified but will we get any value from allowing metadata to be specified that isn't needed by runner?. 

If it is metadata that is used by runner but not needed by the `schema_context`, should it even be specified here in the first place?

dwyeradam: I think this check is valid. The point of this PR is to fail early in runner. If we define metadata we don't then use there could be situations where we fail needlessly in runner.

MebinAbraham:
Closing for now as need to do check for other forms of metadata. ie `"meta":`

PR #50, opened 2018-05-02
Specify required metadata in schema

MebinAbraham:
Runner [PR](https://github.com/ONSdigital/eq-survey-runner/pull/1571) must be merged first. Refer to runner PR for more context.
This PR enforces that:

### How to review
2. Edit a schema to ensure that if you do not define the piped metadata as required then it fails.

Ways of piping metadata:

1. `metadata.ru_name`
2. `metadata['ru_name']`
3. `"meta": "ru_name"`

> Update: Made metadata optional so that this can be merged without affecting (master) runner. The publisher [PR](https://github.com/ONSdigital/eq-publisher/pull/64) can also go in. A new validator PR need be raised to be make metadata mandatory once the runner PR is ready to be merged.

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -26,6 +26,8 @@ class Validator:
 
         errors = []
 
+        # errors.extend(self._validate_schema_contain_metadata(json_to_validate))
+
         numeric_answer_ranges = {}
 
         all_groups = []
@@ -82,6 +84,32 @@ class Validator:
             }
         except SchemaError as e:
             return '{}'.format(e)
+
+    def _validate_schema_contain_metadata(self, schema):
+
+        errors = []
+
+        # user_id and period_id required downstream for receipting
+        # ru_name required for template rendering
+        default_metadata = ['user_id', 'period_id']
+        schema_metadata = schema['metadata']
+
+        # Find all words that precede any of:
+        all_metadata = set(re.findall(r"((?<=metadata\[\')\w+"  # metadata['
+                                      r'|(?<=metadata\.)\w+'  # metadata.
+                                      r"|(?<=meta\': \')\w+)", str(schema)))  # meta': '
+
+        # Checks if piped/routed metadata is defined in the schema
+        for metadata in all_metadata:
+            if metadata not in schema_metadata.keys():
+                errors.append(self._error_message('Metadata - {} not specified in metadata field'.format(metadata)))
+
+        # Checks if unused metadata is defined
+        for metadata in schema_metadata.keys():
+            if metadata not in all_metadata and metadata not in default_metadata:
+                errors.append(self._error_message('Unused metadata defined in metadata field - {}'.format(metadata)))
+
+        return errors
 
     def validate_calculated_ids_in_answers_to_calculate_exists(self, question):
         # Validates that any answer ids within the 'answer_to_group'
--- a/schemas/questionnaire_v1.json
+++ b/schemas/questionnaire_v1.json
@@ -59,6 +59,29 @@
         "StatisticsOfTradeAct"
       ]
     },
+    "metadata": {
+      "type": "object",
+      "required": [
+        "user_id",
+        "period_id"
+      ],
+      "patternProperties": {
+        "^.*$": {
+          "type": "object",
+          "properties": {
+            "validator": {
+              "type": "string",
+              "enum": [
+                "date",
+                "string",
+                "object"
+              ]
+            }
+          }
+        }
+      },
+      "additionalProperties": false
+    },
     "view_submitted_response": {
       "type": "object",
       "properties": {
--- a/tests/schemas/test_invalid_date_range_period.json
+++ b/tests/schemas/test_invalid_date_range_period.json
@@ -7,8 +7,13 @@
     "description": "A test schema for different date formats",
     "theme": "default",
     "legal_basis": "StatisticsOfTradeAct",
-    "variables": {
-        "period": "{{ format_conditional_date (answers.period_from, exercise.start_date)}} to {{ format_conditional_date (answers.period_to, exercise.end_date)}}"
+    "metadata": {
+        "user_id": {
+            "validator": "string"
+        },
+        "period_id": {
+            "validator": "string"
+        }
     },
     "sections": [{
         "id": "default-section",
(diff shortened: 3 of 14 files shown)


app/validation/validator.py
```diff
@@ -86,6 +88,27 @@ def _validate_json_against_schema(self, json_to_validate):
...
             return '{}'.format(e)
 
+    def _validate_schema_contain_metadata(self, schema):
+
+        errors = []
+        default_metadata = ['user_id', 'period_id']
+        schema_metadata = schema['metadata']
+
+        # Find all words that precede with "metadata[' or meta': '"
+        parsed_metadata = re.findall(r"((?<=metadata\[\')\w+|(?<=meta\': \')\w+)", str(schema))
```
shahi645: Just a question for curiosity, is regex the best way to to find all metadata values

nealedj: In this case I think it makes sense. It matches the way that runner treats the whole schema as a template rather than just replacing values within specific values.

MebinAbraham:
@nealedj Spoke with @samiwel and suggested to raise an issue on publisher which can be prioritised when this is approved.

nealedj:
Approved - but this shouldn't be merged until eq-publisher has a merged PR to add this metadata.

tests/schemas/test_invalid_date_range_period.json
```diff
@@ -7,8 +7,11 @@
     "description": "A test schema for different date formats",
     "theme": "default",
     "legal_basis": "StatisticsOfTradeAct",
-    "variables": {
-        "period": "{{ format_conditional_date (answers.period_from, exercise.start_date)}} to {{ format_conditional_date (answers.period_to, exercise.end_date)}}"
+    "metadata": {
+        "trad_as": "string",
+        "user_id": "string",
+        "period_id": "string",
+        "ru_name": "string"
```
samiwel: Am I correct in thinking that `"string"` above is defining the expected type of the metadata and not a dummy metadata value?

MebinAbraham: @samiwel The keys are to define the type of validator runner will use for each type of metadata. For dates it will be `"ref_p_start_date": "date"`. Its not so much the `type` but what the data represents. I can explain it directly if its not clear.

schemas/questionnaire_v1.json
```diff
@@ -59,6 +60,10 @@
         "StatisticsOfTradeAct"
       ]
     },
+    "metadata": {
+      "type": "object",
+      "required": ["period_id", "user_id", "ru_name"]
```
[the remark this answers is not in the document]
MebinAbraham: @samiwel Updated.

tests/schemas/test_invalid_date_range_period.json
samiwel: Okay, I understand. The only comment I have about that is that Author requires specific knowledge about the various types of validators supported by Survey Runner. For now it's fine, we can manage with strings and dates, but longer term it might be nice to explicitly define the available validators in the schema definitions.

e.g.

```
"trad_as": {
  "validator": "string"
}
```

then you can define `validator` to be an enum type of "string", "date", ...

MebinAbraham: FYI: Remove `ru_name` and `trad_as` as it is not just implemented into runner directly.

@nealedj What do you think about Sam's comment. Is it worth changing now if it is need by runner in the future?. It would require runner changes with schema and logic but I think if its going to change, why not implement it properly in the first place?

nealedj: Yep that works.

If you're making other changes already then this seems like a good way to go. It would also allow us to add further keys to metadata if we needed to for future requirements (I.e. `optional`)

schemas/questionnaire_v1.json
```diff
@@ -62,7 +62,26 @@
...
+      "patternProperties": {
+        "^.*$": {
+          "type": "object",
+          "properties": {
+            "validator": {
+              "type": "string",
+              "enum": [
+                "date",
+                "string",
+                "object"
```
samiwel: What does the object validator check for? I'm not saying its wrong just wondering which metadata values would be passed as an object, do you have an example?

app/validation/validator.py
```diff
@@ -83,6 +85,30 @@ def _validate_json_against_schema(self, json_to_validate):
...
+
+        errors = []
+
+        # user_id and period_id required downstream for receipting
+        # ru_name required for template rendering
+        default_metadata = ['user_id', 'period_id']
+        schema_metadata = schema['metadata']
+
+        # Find all words that precede with "metadata['" or "meta': '"
+        parsed_metadata = re.findall(r"((?<=metadata\[\')\w+|(?<=meta\': \')\w+)", str(schema))
```
samiwel: Correct me if I'm wrong, but I don't think this regex will identify uses of metadata where the dot notation is used? Currently Author does not use the square bracket notation when piping answers (we tried it and had issues with encoding and having to resort to backspacing quotes in our JSON output that we'll need to resolve).

For now its not a big deal as we don't pipe metadata currently, but in the interim, might it be worth extending this regex to check for `metadata.key` in addition to `metadata['key']`?

nealedj: Yeah this was changed due to runner referencing answer IDs instead of answer aliases when piping.

There was a desire to keep the piping consistent. I don't see any issue with allowing either form, it's a valid way to construct the jinja template expression. And especially not if author's not in a place to adopt square brackets.

MebinAbraham: I made it only bracket notation as that what I was told we would use. Both notations are valid of course, it was just from a point of consistency. If Author requires dot notation thats fine, I'm happy to modify the regex.

samiwel: For now author requires dot notation. Hopefully we'll address that at some point, but it's low down on our priorities at the moment. BTW, raised a PR on Publisher to add in the metadata for these values.

MebinAbraham: Great, thanks.

MebinAbraham:
@nealedj @samiwel @shahi645 Can you re-review when you get a chance. Thanks

MebinAbraham:
@nealedj @samiwel @shahi645 

> Update: Made metadata optional so that this can be merged without affecting (master) runner. The publisher PR can also go in. A new validator PR need be raised to be make metadata mandatory once the runner PR is ready to be merged.

PR #52, opened 2018-05-21
Eq 1990 different versions of question titles

shahi645:
We have defined a question `titles` object that allows for different versions of a question title depending on a given answer in a previous question. We also have defined the conditional(s) in the titles object to be exactly how when rules work. 

A question can now have a `title` string or `titles` object. Each `titles` object must have a default value, which is always the last value present, this cannot have any conditional with it as its the default value.

We have also added tests so that any answer id's in `when` conditions are valid answer id's.

### How to review 
Run SR branch https://github.com/ONSdigital/eq-survey-runner/tree/eq-1990-different-versions-of-question-titles against this branch, it should pass, and then try and break it based on our rules

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -29,6 +29,7 @@ class Validator:
         errors.extend(self._validate_schema_contain_metadata(json_to_validate))
 
         numeric_answer_ranges = {}
+        all_answer_ids = self._get_all_answer_ids(json_to_validate)
 
         all_groups = []
         for section in json_to_validate.get('sections'):
@@ -52,9 +53,13 @@ class Validator:
                         errors.extend(self.validate_calculated_ids_in_answers_to_calculate_exists(question))
                         errors.extend(self.validate_child_answers_define_parent(question['answers']))
                         errors.extend(self.validate_date_range(question))
+                        if question.get('titles'):
+                            errors.extend(self.validate_multiple_question_titles(question['titles'],
+                                                                                 question['id'],
+                                                                                 all_answer_ids))
 
                         for answer in question['answers']:
-                            errors.extend(self.validate_routing_on_answer_options(block, answer))
+                            errors.extend(self.validate_routing_on_answer_options(block, answer, all_answer_ids))
                             errors.extend(self.validate_duplicate_options(answer))
 
                             if answer['type'] == 'Date':
@@ -140,7 +145,7 @@ class Validator:
 
         return errors
 
-    def validate_routing_on_answer_options(self, block, answer):
+    def validate_routing_on_answer_options(self, block, answer, answer_ids):
         answer_errors = []
         if 'routing_rules' in block and block['routing_rules'] and 'options' in answer:
             options = [option['value'] for option in answer['options']]
@@ -148,9 +153,15 @@ class Validator:
 
             for rule in block['routing_rules']:
                 if 'goto' in rule and 'when' in rule['goto'].keys():
-                    when = rule['goto']['when']
-                    if 'id' in when and when['id'] == answer['id'] and when['value'] in options:
-                        options.remove(when['value'])
+                    when_clause = rule['goto']['when']
+                    is_valid, error_message = self.validate_when_rule(when_clause, answer_ids, block['id'])
+                    if not is_valid:
+                        answer_errors.append(self._error_message(error_message))
+
+                    for when in when_clause:
+                        if 'id' in when and 'value' in when:
+                            if when['id'] == answer['id'] and when['value'] in options:
+                                options.remove(when['value'])
                 else:
                     options = []
                     has_default_route = True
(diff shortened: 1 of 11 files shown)


app/validation/validator.py
```diff
@@ -149,8 +155,12 @@ def validate_routing_on_answer_options(self, block, answer):
             for rule in block['routing_rules']:
                 if 'goto' in rule and 'when' in rule['goto'].keys():
                     when = rule['goto']['when']
-                    if 'id' in when and when['id'] == answer['id'] and when['value'] in options:
-                        options.remove(when['value'])
+                    if 'id' in when[0]:
```
nealedj: Looks like this was already broken. But this is explicitly only check the first `when` rule. Could you iterate through the `when` list and check each rule in it instead of just grabbing the first one?

app/validation/validator.py
```diff
@@ -168,6 +178,35 @@ def validate_routing_on_answer_options(self, block, answer):
                 answer_errors.append(self._error_message(unrouted_error))
         return answer_errors
 
+    @staticmethod
+    def validate_when_rule(when_clause, existing_ids):
+        # Validates any answer id in a when clause exists within the schema
+        when_clause_ids = [when_clause[0]['id']]
```
nealedj: It looks like this is expecting to extract an answer ID from each when clause. But then just creates an array with a single item.

See my comment about iterating through all when rules.

app/validation/validator.py
```diff
@@ -168,6 +178,35 @@ def validate_routing_on_answer_options(self, block, answer):
...
+
+        for answer_id in when_clause_ids:
```
nealedj: Could also be checked as:

```
return set(when_clause_ids).difference(existing_ids) == 0
```

app/validation/validator.py
```diff
@@ -168,6 +178,35 @@ def validate_routing_on_answer_options(self, block, answer):
                 answer_errors.append(self._error_message(unrouted_error))
         return answer_errors
 
+    @staticmethod
+    def validate_when_rule(when_clause, existing_ids):
```
nealedj: `validate_when_rule` implies that it'll evaluate any `when` rule. But actually this method will error if you pass any `when` rule it at that doesn't contain an `id` attribute.

I'd suggest either checking that `id` is in each rule as part of this method or calling this method something like `validate_answer_id_when_rule`. I prefer the former.

app/validation/validator.py
```diff
@@ -149,8 +155,12 @@ def validate_routing_on_answer_options(self, block, answer):
...
                 if 'goto' in rule and 'when' in rule['goto'].keys():
                     when = rule['goto']['when']
-                    if 'id' in when and when['id'] == answer['id'] and when['value'] in options:
-                        options.remove(when['value'])
+                    if 'id' in when[0]:
+                        if not self.validate_when_rule(when, answer_ids):
+                            answer_errors.append(self._error_message('The answer id - {} in the "when" clause for {} '
+                                                                     'does not exist'.format(when[0]['id'], block['id'])))
+                        if when[0]['id'] == answer['id'] and when[0]['value'] in options:
+                            options.remove(when[0]['value'])
```
nealedj: I know this wasn't introduced in this change. But what is this bit of code doing? Why is it removing the value from the answer options if it's used in the routing rules. If you know, could you add a comment please?

app/validation/validator.py
nealedj: Good call on adding this check though 👍

app/validation/validator.py
```diff
@@ -168,6 +178,35 @@ def validate_routing_on_answer_options(self, block, answer):
...
+        when_clause_ids = [when_clause[0]['id']]
+
+        for answer_id in when_clause_ids:
+            if answer_id not in existing_ids:
+                return False
+
+        return True
+
+    def validate_multiple_question_titles(self, question_titles, question_id, answer_ids):
+        # Validates that the last title in a question titles object contains only a value key
```
nealedj: Useful comment 👍

app/validation/validator.py
```diff
@@ -168,6 +178,35 @@ def validate_routing_on_answer_options(self, block, answer):
...
+        errors = []
+
+        last_title = question_titles[-1]
+        if len(last_title) != 1 or 'value' not in last_title:
+            errors.append(self._error_message('The last value must be the default value with no "when" clause for {}'
+                                              .format(question_id)))
+
+        for title in question_titles:
+            if 'when' in title:
+                if not self.validate_when_rule(title['when'], answer_ids):
```
nealedj: Won't this fail if the `when` rule doesn't contain an `id`?

shahi645:
@nealedj Thats strange, wasnt failing for me. I will take a look shortly

app/validation/validator.py
shahi645: @nealedj For some reason it checks answers options, and takes away any values that match in when rule, the result then is used to evaluate if there are unrouted options in the answer see line `170`. It seems a bit hacky to be honest. Maybe tomorrow we could go over it and change appropriately?

shahi645: I think thats why you were seeing those errors, also i dont think this takes into account default routing, i.e. when it just routes if no other `when` is evaluated true

app/validation/validator.py
```diff
@@ -154,13 +154,15 @@ def validate_routing_on_answer_options(self, block, answer, answer_ids):
...
                 if 'goto' in rule and 'when' in rule['goto'].keys():
-                    when = rule['goto']['when']
-                    if 'id' in when[0]:
-                        if not self.validate_when_rule(when, answer_ids):
-                            answer_errors.append(self._error_message('The answer id - {} in the "when" clause for {} '
-                                                                     'does not exist'.format(when[0]['id'], block['id'])))
-                        if when[0]['id'] == answer['id'] and when[0]['value'] in options:
-                            options.remove(when[0]['value'])
+                    when_clause = rule['goto']['when']
+                    valid_when_clause = self.validate_when_rule(when_clause, answer_ids, block['id'])
```
nealedj: Nitpick - since this method returns a tuple I'd suggest unpacking it for clarity rather than having to index it.

I.e.
```
is_valid, error_message = self.validate_when_rule(when_clause, answer_ids, block['id'])
if not is_valid:
   answer_errors.append(self._error_message(error_message))
```

app/validation/validator.py
```diff
@@ -179,15 +181,16 @@ def validate_routing_on_answer_options(self, block, answer, answer_ids):
         return answer_errors
 
     @staticmethod
-    def validate_when_rule(when_clause, existing_ids):
+    def validate_when_rule(when_clause, existing_ids, referenced_id):
         # Validates any answer id in a when clause exists within the schema
-        when_clause_ids = [when_clause[0]['id']]
+        for when in when_clause:
+            if 'id' in when:
+                message = 'The answer id - {} in the "when" clause for {} does not exist'.format(when['id'],
```
nealedj: Nitpick - consider building the `message` inside the following `if` condition since it's only relevant within that scope.

app/validation/validator.py
shahi645: very good point 👍

nealedj:
Actually disregard that.

The validator already needs a refactor to fix the `too-many-branches` error. Not going to block this work on the need for a general refactor.

nealedj:
@shahi645 will merge once these 2 commits are squished.

PR #53, opened 2018-05-30
Enforce metadata in schemas

MebinAbraham:
Runner [PR](https://github.com/ONSdigital/eq-survey-runner/pull/1571) must be merged first. Refer to runner PR for more context.
This PR enforces that:

### How to review
2. Edit a schema to ensure that if you do not define the piped metadata as required then it fails and vice versa.

Ways of piping metadata:

1. `metadata.ru_name`
2. `metadata['ru_name']`
3. `"meta": "ru_name"`

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -26,7 +26,7 @@ class Validator:
 
         errors = []
 
-        # errors.extend(self._validate_schema_contain_metadata(json_to_validate))
+        errors.extend(self._validate_schema_contain_metadata(json_to_validate))
 
         numeric_answer_ranges = {}
 
--- a/schemas/questionnaire_v1.json
+++ b/schemas/questionnaire_v1.json
@@ -9,7 +9,8 @@
     "title",
     "sections",
     "theme",
-    "legal_basis"
+    "legal_basis",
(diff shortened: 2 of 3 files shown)


app/validation/validator.py
```diff
@@ -26,7 +26,7 @@ def validate_schema(self, json_to_validate):  # noqa: C901  pylint: disable=too-
 
         errors = []
 
-        # errors.extend(self._validate_schema_contain_metadata(json_to_validate))
+        errors.extend(self._validate_schema_contain_metadata(json_to_validate))
```
nealedj: We're in an odd position now. Author _should_ be failing to preview questionnaires with piped values due to this. However, because author still uses dot notation instead of square brackets to pipe values it's still working.

I think we should allow both dot notation and square brackets - and we should validate both.

@samiwel FYI - any thoughts?

MebinAbraham: @nealedj You can pipe using both dot or bracket notation. I've already made it to allow both.

nealedj: Sorry I was being daft and getting this mixed up with piped answers. Ignore me!

PR #54, opened 2018-05-30
Add MutuallyExclusiveCheckbox answer type

dwyeradam:
Format same as standard checkbox.

### DIFF
--- a/schemas/answers/checkbox.json
+++ b/schemas/answers/checkbox.json
@@ -21,7 +21,8 @@
       "type": {
         "type": "string",
         "enum": [
-          "Checkbox"
+          "Checkbox",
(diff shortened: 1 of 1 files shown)


shahi645:
@dwyeradam Aren't we looking to add some validator rules, regarding the last option being the `OR` value, or is that out of the scope of `schema-validator`?

dwyeradam:
@shahi645, schema-validator couldn't possibly know which the OR option is as it's not defined in any other way than just being last? This was the approach decided in the tech session but we could review wether some kind of OR tag is required. I think that may be overkill though as it's obvious enough when you preview a schema that OR is last.

shahi645:
I understand in which case ill just run your branch against this to test, but looks fine to me.

PR #56, opened 2018-06-07
Add validation for white space in survey id

warrenbailey:
Add validation rule for white space in survey id. Also added a unti test and updated the run_test.sh to use pipenv to run pytest.

### DIFF
--- a/scripts/run_tests_unit.sh
+++ b/scripts/run_tests_unit.sh
@@ -19,5 +19,5 @@ function display_result {
   fi
 }
 
-py.test "$@"
+pipenv run py.test "$@"
 display_result $? 3 "Unit tests"
--- a/tests/schemas/test_invalid_survey_id_whitespace.json
+++ b/tests/schemas/test_invalid_survey_id_whitespace.json
@@ -1,0 +1,3364 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
+    "survey_id": "lms ",
+    "title": "Labour Market Survey (Alpha)",
+    "description": "Labour Market Survey Schema (Alpha)",
+    "theme": "default",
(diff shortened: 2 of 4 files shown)


scripts/run_tests_unit.sh
```diff
@@ -19,5 +19,5 @@ function display_result {
   fi
 }
 
-py.test "$@"
+pipenv run py.test "$@"
```
shahi645: @warrenbailey we just use `pipenv run` in the terminal normally, but I suppose it makes no difference

tests/schemas/test_invalid_survey_id_whitespace.json
```diff
@@ -0,0 +1,3364 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
+    "survey_id": "lms ",
+    "title": "Labour Market Survey (Alpha)",
```
shahi645: we don't need such a big test_schema for this test, again i suppose it makes no difference

tests/test_schema_validation.py
```diff
@@ -188,6 +188,15 @@ def test_invalid_question_titles_object(self):
...
                                                '"when" clause for what-gender-question does not exist')
 
+    def test_invalid_survey_id_whitespace(self):
+
+        file = 'schemas/test_invalid_survey_id_whitespace.json'
+        json_to_validate = self.open_and_load_schema_file(file)
+
+        errors = self.validator.validate_schema(json_to_validate)
+
+        self.assertEquals(errors.get('message'), "'lms ' does not match '^[0-9a-z]+$'")
```
shahi645: Where does the message come from in this instance? I think it could be changed to something more readable like "Please remove all whitespaces from survey_id 'lms '"

PR #57, opened 2018-06-07
Add YearDate answer type

dwyeradam:
New YearDate answer type.
Min/Max and Range offsets can only specify years.

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -234,11 +234,19 @@ class Validator:
                                                       .format(question['id'])))
 
             first_answer_type = question['answers'][0]['type']
-            if (first_answer_type == 'MonthYearDate'
-                    and ('minimum' in period_limits and 'days' in period_limits['minimum']
-                         or 'maximum' in period_limits and 'days' in period_limits['maximum'])):
+
+            has_days_limit = 'days' in period_limits.get('minimum', []) \
+                             or 'days' in period_limits.get('maximum', [])
+            has_months_limit = 'months' in period_limits.get('minimum', []) \
+                               or 'months' in period_limits.get('maximum', [])
+
+            if first_answer_type == 'MonthYearDate' and has_days_limit:
                 errors.append(self._error_message('Days can not be used in period_limit for yyyy-mm date range for {}'
                                                   .format(question['id'])))
+
+            if first_answer_type == 'YearDate' and (has_days_limit or has_months_limit):
+                errors.append(self._error_message('Days/Months can not be used in period_limit for yyyy date range'
+                                                  ' for {}'.format(question['id'])))
 
         return errors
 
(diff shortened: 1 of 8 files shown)


tests/test_schema_validation.py
```diff
@@ -149,6 +149,18 @@ def test_invalid_mm_yyyy_date_range_period(self):
...
+        json_to_validate = self.open_and_load_schema_file(file)
+
+        errors = self.validator.validate_schema(json_to_validate)
+
+        self.assertEqual(len(errors), 1)
+        self.assertEqual(errors[0]['message'], 'Schema Integrity Error. Days/Months can not be used in period_limit'
+                                               ' for yyyy date range for date-range-question')
+
+
     def test_invalid_single_date_period(self):
```
MebinAbraham: Consistency wise. Too many blank lines.

app/validation/validator.py
```diff
@@ -240,6 +240,14 @@ def validate_date_range(self, question):
...
                                                   .format(question['id'])))
 
+            if (first_answer_type == 'YearDate'
+                    and ('minimum' in period_limits and ('days' in period_limits['minimum']
+                                                         or 'months' in period_limits['minimum'])
+                         or 'maximum' in period_limits and ('days' in period_limits['maximum']
+                                                            or 'months' in period_limits['maximum']))):
+                errors.append(self._error_message('Days/Months can not be used in period_limit for yyyy date range'
+                                                  ' for {}'.format(question['id'])))
+
```
MebinAbraham: Also `R:243,16: Too many boolean expressions in if statement (7/5) (too-many-boolean-expressions)`

dwyeradam: I can split this into two separate statements. One for Days and one for Months but that looks messy and could return twice as many errors as necessary. I'll see if I can reduce the boolean expressions but keep it simple to understand. I didn't want a complex "If" that's actually harder to understand but has less expressions.

nealedj: Possible suggestion:

```
if first_answer_type == 'YearDate':
    has_min_limit = 'days' in period_limits['minimum'] or 'months' in period_limits['minimum']
    has_max_limit = 'days' in period_limits['maximum'] or 'months' in period_limits['maximum']
    if has_min_limit or has_max_limit:
        errors.append(self._error_message('Days/Months can not be used in period_limit for yyyy date range'
                                        ' for {}'.format(question['id'])))
```

dwyeradam: As above but:

dwyeradam: This can actually then go above and be used for mm-yyyy type as well.

nealedj: Nice 👏

shahi645: Bravo 👍

tests/test_schema_validation.py
dwyeradam: So petty but done.

PR #59, opened 2018-07-04
removed group_instance from when

AndrewTorrance:
Removed the ability to specify a group instance from a when clause . 
The titles work had introduced the  ability to specify a specific group instance to use as a conditional for a when . However in this work it was realised that would only ever be zero in practice , and that it was only needed for a conditional within a repeating group dependant on an answer outside of a repeating group. Therefore runner changed to detect this , and hence group instance not needed in validator


MebinAbraham:
So was this introduced but never got used in the schemas?

dwyeradam:
This is in reference to https://github.com/ONSdigital/eq-survey-runner/pull/1631
Group instance in routing rules was removed in favour of new code that evaluated if you were trying to route from within a repeating group using a value from outside of that routing group.

PR #62, opened 2018-07-13
Add a Duration answer type

kierenbeckett:
Add a Duration type.  Uses a field named `units` to define what the duration is made up of, currently only years/months are supported.


AndrewTorrance:
Could you add a description for the PR please .

kierenbeckett:
Presumably adding it to `Calculated` requires changes to runner, so we should only add it into the schema once there's support for it?

PR #63, opened 2018-07-13
validating id in when of routing rules is a valid answer

AndrewTorrance:
Armed forces household census was throwing 500s, it had a when dependency on an id that does not exist , this was being missed by validator.
The root cause was that the id in the when was only being checked if an answer type was options. 
Code changed so as to validate that the independent variable is a known answer.

To Test:
Test schema changed to have invalid id for a non options answer type, so just run tests.

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -45,8 +45,10 @@ class Validator:
                         errors.extend(self.validate_contains_confirmation_or_summary(block))
 
                     for rule in block.get('routing_rules', []):
-                        errors.extend(self.validate_schema_contains_valid_routing_rules(group['blocks'], 'block', rule))
-                        errors.extend(self.validate_schema_contains_valid_routing_rules(all_groups, 'group', rule))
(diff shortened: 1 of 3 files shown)


app/validation/validator.py
```diff
@@ -190,10 +198,8 @@ def validate_when_rule(when_clause, existing_ids, referenced_id):
         for when in when_clause:
             if 'id' in when:
                 if when['id'] not in existing_ids:
-                    return False, 'The answer id - {} in the "when" clause for {} does not exist'.format(when['id'],
-                                                                                                         referenced_id)
-
-        return True, 'Valid'
+                    return Validator._error_message('The answer id - {} in the "when" clause for {} does not exist'.format(when['id'],
+                                                                                                         referenced_id))
```
nealedj: Linting error on this line:

```
./app/validation/validator.py:202:106: E128 continuation line under-indented for visual indent
```

MebinAbraham:
We should also address this for `skip_conditions`.

app/validation/validator.py
AndrewTorrance: Done

PR #65, opened 2018-07-16
Return a "best match" reason for errors

benhowes:
Uses [best match](https://python-jsonschema.readthedocs.io/en/v2.6.0/errors/#best-match-and-relevance) from the python jsonschema library to return a best guess reason in addition to the erroneous block.


jonnyshaw89:
@benhowes you need to setup signed commits for us to be able to merge this.

PR #67, opened 2018-07-23
Add a repeating type of 'until'

kierenbeckett:
Under `routing_rules` in `repeat` we want to support a new type of `until` that will repeat until a condition is met (specified with a `when` clause).  The when clause only supports a subset of the behaviours it does elsewhere, only one is allowed in the list and only on `id`.

### DIFF
--- a/api.py
+++ b/api.py
@@ -8,4 +8,4 @@ application.register_blueprint(validate_blueprint)
 application.register_blueprint(status_blueprint)
 
 if __name__ == '__main__':
-    application.run()
+    application.run(port=5001)
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -30,7 +30,7 @@ class Validator:
         errors.extend(self._validate_schema_contain_metadata(json_to_validate))
 
         numeric_answer_ranges = {}
-        all_answer_ids = self._get_all_answer_ids(json_to_validate)
+        answer_ids_with_group_id = self._get_answer_ids_with_group_id(json_to_validate)
 
         all_groups = []
         for section in json_to_validate.get('sections'):
@@ -38,6 +38,14 @@ class Validator:
 
         for section in json_to_validate['sections']:
             for group in section['groups']:
+
+                for rule in group.get('routing_rules', []):
+                    errors.extend(self.validate_schema_routing_rule_routes_to_valid_target(group['blocks'], 'block', rule))
+                    errors.extend(self.validate_schema_routing_rule_routes_to_valid_target(all_groups, 'group', rule))
+
+                    errors.extend(self.validate_schema_routing_rule_dependent_on_valid_answer(rule, answer_ids_with_group_id, group))
+                    errors.extend(self.validate_repeat_when_rule_restricted(rule, answer_ids_with_group_id, group))
+
                 for block in group['blocks']:
 
                     if section == json_to_validate['sections'][-1] \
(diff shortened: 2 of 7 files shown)


api.py
```diff
@@ -8,4 +8,4 @@
 application.register_blueprint(status_blueprint)
 
 if __name__ == '__main__':
-    application.run()
+    application.run(port=5001)
```
jonnyshaw89: Wont this break the docker container?

schemas/common_definitions.json
```diff
@@ -349,24 +350,100 @@
...
+          "oneOf": [
+              {
+                "type": "object",
+                "properties": {
+                  "type": {
+                    "enum": [
+                      "until"
+                    ]
+                  },
+                  "when": {
```
jonnyshaw89: Can we not use the existing `when` definition to reduce duplication?
`"$ref": "common_definitions.json#/when"`

api.py
kierenbeckett: Docker uses `flask run`

schemas/common_definitions.json
kierenbeckett: I'd have liked to, but the `when` in repeats only supports a subset of the usual behaviour, only one condition is allowed and only based on an `id`.

jonnyshaw89: my concern is that we will have 2 definitions of `when` and they could then deviate from each other and it would get confusing.

kierenbeckett: what do you suggest instead?

api.py
insacuri: Shouldn't this be done via a setting?  Seems odd to me to hard code it here

kierenbeckett: This line exists purely for running locally, not sure if needs to be a setting if it doesn't change per environment.

schemas/common_definitions.json
jonnyshaw89: I think it would be better to reuse the existing definition but then limit the usage of it in `validator.py`

app/validation/validator.py
```diff
@@ -581,14 +621,14 @@ def convert_to_datetime(value):
...
+        answer_ids = {}
 
         for section in json_to_validate.get('sections'):
             for group in section.get('groups'):
                 for block in group.get('blocks'):
                     if block.get('questions'):
                         for question in block['questions']:
                             for answer in question.get('answers', []):
-                                answer_ids.append(answer['id'])
+                                answer_ids[answer['id']] = {'group_id': group['id']}
```
jonnyshaw89: This has drastically changed the behaviour of this function so we should probably rename it?

jonnyshaw89: I also can't see why this change was made?

kierenbeckett: Not really, still returns answer IDs, just with some metadata attached.  We pass around the group IDs so in repeat rules we can check the answer is in the same group.

jonnyshaw89: it feels like a bit of a hack just to get the group id for an answer in 1 function?
I don't think that the function name describes what is returned.

app/validation/validator.py
```diff
@@ -580,15 +620,15 @@ def convert_to_datetime(value):
...
+        answers = {}
 
         for section in json_to_validate.get('sections'):
             for group in section.get('groups'):
                 for block in group.get('blocks'):
                     if block.get('questions'):
                         for question in block['questions']:
                             for answer in question.get('answers', []):
-                                answer_ids.append(answer['id'])
+                                answers[answer['id']] = {'group_id': group['id']}
```
jonnyshaw89: given we now just call this `_get_all_answers` should we not make this `answers[answer['id']] = answer` ?

kierenbeckett: Answers don't contain a reference to their group

jonnyshaw89: then i think the name is more confusing as its not getting answers?

kierenbeckett:
🤣

PR #69, opened 2018-07-25
Add CalculatedSummary block type

dwyeradam:
Json schema checks
Answers used in calculation must be of same type (answer type, unit, currency)

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -18,7 +18,7 @@ class Validator:
         with open('schemas/questionnaire_v1.json', encoding='utf8') as schema_data:
             self.schema = load(schema_data)
 
-    def validate_schema(self, json_to_validate):  # noqa: C901  pylint: disable=too-complex
+    def validate_schema(self, json_to_validate):  # noqa: C901  pylint: disable=too-complex, too-many-branches
 
         schema_errors = self._validate_json_against_schema(json_to_validate)
 
@@ -30,7 +30,7 @@ class Validator:
         errors.extend(self._validate_schema_contain_metadata(json_to_validate))
 
         numeric_answer_ranges = {}
-        answer_ids_with_group_id = self._get_answer_ids_with_group_id(json_to_validate)
+        answers_with_parent_ids = self._get_answers_with_parent_ids(json_to_validate)
 
         all_groups = []
         for section in json_to_validate.get('sections'):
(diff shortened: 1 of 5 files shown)


app/validation/validator.py
```diff
@@ -204,6 +207,38 @@ def validate_repeat_rule_restricted(rule, block):
...
         return errors
 
+
+    def validate_calculated_summary_type(self, block, answers_with_parent_ids):
+        answer_types = [
+            answers_with_parent_ids[answer_id]['answer']['type']
+            for answer_id in block['calculation']['answers_to_calculate']
+        ]
+
+        if not all(type == answer_types[0] for type in answer_types):
```
benhowes: Would be best not to override the builtin `type` variable here and line 227, 236

schemas/blocks/calculated_summary.json
```diff
@@ -0,0 +1,61 @@
...
+        "properties": {
+          "calculation_type": {
+            "type": "string",
+            "enum": [
+              "sum"
+            ]
+          },
+          "answers_to_calculate": {
+            "type": "array",
+            "minItems": 2,
```
benhowes: In the scenario that you're testing a survey part way through building it, is it conceivable that you'd run a survey where you've used a calculate block, but only have a single answer id in this block at the point you want to run a test of your other logic? My point is that it's probably a bit silly to only have one item in `answers_to_calculate`, but perhaps it shouldn't be invalid?

app/validation/validator.py
dwyeradam: Yes, stupid of me, will call it answer_type.

schemas/blocks/calculated_summary.json
dwyeradam: I agree in that situation you may want just one answer but I believe schema validator is to protect against poorly designed schemas and to do that we have to treat each survey as complete at the point of validation. If you only need one answer then you should use the Confirmation block type and change it after.

benhowes: Okay @dwyeradam, I'll back off on that. I do wonder if having the ability to raise "Warnings/Recommendations" as well as hard errors may be useful in future?

dwyeradam: @benhowes Only back off if you're happy with my response. Otherwise we open it up to the group and hope for some consensus. Often with varying success. I won't take it personally.

MebinAbraham:
Ideally would like to test against the runner changes.

dwyeradam:
@benhowes did you want a quick look at the latest two commits. Just two additional tests. One to check for invalid answer_ids and another to test for duplicate answer ids.

benhowes:
@dwyeradam Looks good to me! I should've thought of the duplicate id's thing too!

PR #71, opened 2018-07-27
Add a repeat type of 'group'

kierenbeckett:
Add a repeat type of 'group'

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -43,7 +43,7 @@ class Validator:
                     errors.extend(self.validate_schema_routing_rule_routes_to_valid_target(group['blocks'], 'block', rule))
                     errors.extend(self.validate_schema_routing_rule_routes_to_valid_target(all_groups, 'group', rule))
 
-                    errors.extend(self.validate_schema_routing_rule_dependent_on_valid_answer(rule, answers_with_parent_ids, group))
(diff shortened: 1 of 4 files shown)


app/validation/validator.py
```diff
@@ -165,13 +165,23 @@ def validate_schema_routing_rule_routes_to_valid_target(self, dict_list, goto_ke
...
+    def validate_routing_rule(self, rule, all_groups, answer_ids_with_group_id, block_or_group):
         errors = []
 
-        rule = rule.get('goto') or rule.get('repeat')
-        if 'when' in rule:
-            when_errors = Validator.validate_when_rule(rule['when'], answer_ids_with_group_id, block_or_group['id'])
+        when = None
+        if 'goto' in rule:
+            when = rule['goto'].get('when')
+        elif 'repeat' in rule:
```
benhowes: Something about this block gives me the feeling it wont be very readable in a few months. Perhaps a method for `validate_repeat_over_group_ids` could be added, then you can test for that special case and return early from this method?

kierenbeckett:
It's in a separate block as it has different properties.  It's failing because of the error in the schema in survey runner, please can you update survey runner and try again.

MebinAbraham:
@kierenbeckett Can squash. Worth waiting on merge until the runner PR is approved?

PR #73, opened 2018-08-02
Add boolean metadata value type, remove object

benhowes:
In aid of flattening out `variant_flags` in to the top level claims object of the JWT, it became apparent that a boolean validator type was needed.

Tested against the updated schemas in https://github.com/ONSdigital/eq-survey-runner/pull/1715

### DIFF
--- a/schemas/questionnaire_v1.json
+++ b/schemas/questionnaire_v1.json
@@ -72,9 +72,10 @@
           "validator": {
             "type": "string",
             "enum": [
+              "boolean",
               "date",
               "string",
-              "object"
+              "uuid"
             ]
           }
         },
--- a/tests/schemas/test_invalid_metadata.json
+++ b/tests/schemas/test_invalid_metadata.json
@@ -44,12 +44,16 @@
       "validator": "date"
     },
     {
-      "name": "variant_flags",
-      "validator": "object"
+      "name": "flag",
+      "validator": "boolean"
     },
     {
       "name": "period_str",
       "validator": "string"
+    },
+    {
+      "name": "case_id",
+      "validator": "uuid"
     }
   ],
   "mime_type": "application/json/ons/eq",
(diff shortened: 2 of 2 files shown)


schemas/questionnaire_v1.json
```diff
@@ -72,9 +72,9 @@
           "validator": {
             "type": "string",
             "enum": [
+              "boolean",
               "date",
-              "string",
-              "object"
+              "string"
```
jonnyshaw89: While you are in this code we have this validator defined that isn't allow in through here
https://github.com/ONSdigital/eq-survey-runner/pull/1715/files#diff-f8f1a812b2bb69e21c3e107b2626e63eR39

benhowes: Sorry, are you suggesting to allow it through, or remove the UUID type from runner?

jonnyshaw89: Allow I guess. Either that or remove the code? 
Maybe remove as it’s obviously not used?
Are any of the items currently marked as string actually uuid, if so do we actually care if they aren’t uuid?

benhowes: The only thing documented as being a UUID is `case_id`, however that's not actually validated because it's optional and just passed through as a string. 

I did actually write unit tests for the UUID validation yesterday because I needed to write unit tests for the new boolean validator, so we can be confident that the validation works. We may as well leave it in given that it's a general purpose product?

jonnyshaw89:
@benhowes can you squash these commits?

benhowes:
@jonnyshaw89 @MebinAbraham I've added the UUID type here too to match the runner, all in the latest commit.

jonnyshaw89:
👍

benhowes:
As @dcdarrell9 mentioned in https://github.com/ONSdigital/eq-survey-runner/pull/1715#issuecomment-410189294, merging this prior to https://github.com/ONSdigital/go-launch-a-survey/pull/111 will mean that we can't launch surveys with `variant_flags`, so it's probably best to hold off on this until the survey launcher PR is in, however when it is, this should be done ASAP, because the new launcher doesn't support the object types!

jonnyshaw89:
@benhowes Im happy for this to go in before the launcher change.
Nothing is currently using these schemas so I cant see the benefit in leaving these PRs open?

PR #75, opened 2018-08-06
Allow numbers in metadata keys

insacuri:
Loosen restrictions brought about by https://github.com/ONSdigital/eq-schema-validator/pull/74


schemas/questionnaire_v1.json
```diff
@@ -68,7 +68,7 @@
         "properties": {
           "name": {
             "type": "string",
-            "pattern": "^[a-zA-Z-_]+$"
+            "pattern": "^[a-zA-Z_][a-zA-Z0-9-_]+?$"
```
jonnyshaw89: whats the benefit of the `?`?
also why do we have duplication?
should we remove `[a-zA-Z_]`

insacuri:
tested it locally against runner, works fine

MebinAbraham:
@insacuri Squash please.

schemas/questionnaire_v1.json
benhowes: Just for reference, the reason I wrote it like this is so that names don't start with 0-9 or `-`, which might cause problems with some places you use metadata keys. The 2 `[]` groups are similar, but not duplicate. 

The reason then for having `+?` is because we never said we'd disallow 1 char names.

PR #78, opened 2018-08-23
Update legal basis

dcdarrell9:
- Changed legal basis to be non mandatory and a string

- https://github.com/ONSdigital/eq-survey-runner/pull/1748

### DIFF
--- a/schemas/questionnaire_v1.json
+++ b/schemas/questionnaire_v1.json
@@ -9,7 +9,6 @@
     "title",
     "sections",
     "theme",
-    "legal_basis",
     "metadata"
   ],
   "properties": {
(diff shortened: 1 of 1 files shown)


jonnyshaw89:
@dcdarrell9 I think there might be a related Author change with this so chat to @samiwel

samiwel:
@dcdarrell9  @jonnyshaw89 Looking at the change, I think we should be okay. We currently default the value to the string "StatisticsOfTradeAct" but we don't expose this to the users via the UI yet. And we don't provide them a way of changing this via the UI.

As long as it's optional should be fine, but all schemas generated from Author will have a default value until we make a change at our end.

PR #79, opened 2018-08-23
Don't match collection_metadata in validation

bitdivision:
Existing test checks that this works.

https://github.com/ONSdigital/eq-survey-runner/pull/1732

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -130,8 +130,8 @@ class Validator:
             default_metadata.append('ru_name')
 
         # Find all words that precede any of:
-        all_metadata = set(re.findall(r"((?<=metadata\[\')\w+"  # metadata['
-                                      r'|(?<=metadata\.)\w+'  # metadata.
+        all_metadata = set(re.findall(r"((?<!collection_metadata\[\')(?<=metadata\[\')\w+"  # metadata[' not _metadata['
+                                      r'|(?<!collection_metadata\.)(?<=metadata\.)\w+'  # metadata. not _metadata.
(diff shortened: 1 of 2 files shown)


app/validation/validator.py
```diff
@@ -130,8 +130,8 @@ def _validate_schema_contain_metadata(self, schema):
             default_metadata.append('ru_name')
 
         # Find all words that precede any of:
-        all_metadata = set(re.findall(r"((?<=metadata\[\')\w+"  # metadata['
-                                      r'|(?<=metadata\.)\w+'  # metadata.
+        all_metadata = set(re.findall(r"((?<!_metadata\[\')(?<=metadata\[\')\w+"  # metadata[' not _metadata['
```
jonnyshaw89: should we be more specific here and explicitly define `collection_metadata` not just `_metadata`?

bitdivision: I left it a little less strict because I couldn't really see a situation where we would be using `_metadata` and referring to actual metadata. 

I suppose it's possible that someone could set a variable and call it `something_metadata`.

I'll make the change.

bitdivision: Actually, if someone were to define `something_metadata`, I would think they still wouldn't want validation on it.

@jonnyshaw89 - Did you have any situations in mind where this would cause a problem?

jonnyshaw89: When it comes to something like this i always go with explicit rather than implicit. so explicitly target what you are after rather than throwing a broad net

jonnyshaw89: If someone declares `something_metadata` then they will have to make the conscious decision to exclude it.

PR #83, opened 2018-09-03
Changes for comparison_id in when clauses

bitdivision:
Runner is adding a `comparison_id` to the when clause.

This PR updates the json schema to allow for `comparison_id`.

- Since these when conditions can also be used in skip conditions, I have added validation for those too.
- Adding skip condition validation raised an issue with MWS survey in runner.
- Disallow `set` and `unset` with comparison_id
- Disallow comparison_id being used in `when` clauses in titles (since this could cause other issues and isn't needed by a customer)

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -46,6 +46,9 @@ class Validator:
                     errors.extend(self.validate_routing_rule(rule, answers_with_parent_ids, group))
                     errors.extend(self.validate_repeat_when_rule_restricted(rule, answers_with_parent_ids, group))
 
+                for skip_condition in group.get('skip_conditions', []):
+                    errors.extend(self.validate_skip_condition(skip_condition, answers_with_parent_ids, group))
+
                 for block in group['blocks']:
 
                     if section == json_to_validate['sections'][-1] \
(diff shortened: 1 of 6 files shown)


app/validation/validator.py
```diff
@@ -186,6 +192,25 @@ def validate_routing_rule(rule, answer_ids_with_group_id, block_or_group):
...
+    def validate_skip_condition(self, skip_condition, answer_ids_with_group_id, block_or_group):
+        errors = []
+        when = skip_condition.get('when')
+        when_errors = self.validate_when_rule(when, answer_ids_with_group_id, block_or_group['id'])
+        if when_errors:
+            errors.append(when_errors)
+
+        return errors
+
+    def validate_repeat_group_rule(self, rule, all_groups):
```
jonnyshaw89: Is this used?

bitdivision: Yikes. No

tests/schemas/test_invalid_answer_comparison_types.json
```diff
@@ -0,0 +1,263 @@
...
+                "type": "Interstitial",
+                "id": "equals-answers",
+                "title": "Second equal first",
+                "description": "Your second number was equal to your first number",
+                "skip_conditions": [{
+                    "when": [{
+                        "id": "comparison-1-answer",
+                        "condition": "not equals",
+                        "comparison_id": "comparison-2-answer",
+                        "value": "1"
```
jonnyshaw89: Do we need `"value": "1"`?

PR #88, opened 2018-09-21
Update fields required for unit field

insacuri:
Relates to https://github.com/ONSdigital/eq-survey-runner/pull/1794


README.md
```diff
@@ -19,13 +19,15 @@ pipenv run ./scripts/run_app.sh
...
 
 ```
-pipenv run ./scripts/run_app.sh
+pipenv run ./scripts/run_tests_unit.sh
 ```
 
-then, in another terminal window/tab, navigate to a checked out copy of eq-survey-runner:
+To test the apps functionality:
+
+In another terminal window/tab, navigate to a checked out copy of eq-survey-runner:
```
jonnyshaw89: This no longer makes sense?
Why have we removed the instruction on how to run the app? should we not have added this as well as?

insacuri: I figured the sensible person would know that to test the apps functionality, you need to have the app running.
But fair enough, I like when READMEs are nice and explicit so I'll add it again

jonnyshaw89: Apologies, i didn't realise that above it detailed how to run the app.
In the small diff i saw on here it looked as though the instructions had been removed.

schemas/answers/unit.json
```diff
@@ -85,7 +91,8 @@
       "id",
       "type",
       "mandatory",
-      "unit"
+      "length",
+      "use_plural_for_label"
```
dcdarrell9: I think these two could be `unit_length` & `unit_plural` as they are both related to unit, but not bothered either way 👍

schemas/answers/unit.json
```diff
@@ -45,6 +45,9 @@
       "unit": {
         "$ref": "../lists/units.json#/unit"
       },
+      "unit_length": {
+        "type": "string"
```
bitdivision: I don't know how strict we're being about this, but it's probably worth adding a description to this.

schemas/answers/unit.json
```diff
@@ -85,7 +88,7 @@
       "id",
       "type",
       "mandatory",
-      "unit"
```
bitdivision: Why was this removed?

schemas/answers/unit.json
jonnyshaw89: yeh we should probably make this an enum given there are a fixed set of values

bitdivision: Also, it's probably worth adding:
`"enum": ["short", "narrow", "long"]`

schemas/answers/unit.json
insacuri: Was done mistakenly, readding it.

schemas/answers/unit.json
insacuri: Good point, didn't realise there was an enum field.  Added it now

PR #91, opened 2018-09-27
Make title and label required for AnswerSummary

jonnyshaw89:
Required for https://github.com/ONSdigital/eq-survey-runner/pull/1793

### DIFF
--- a/schemas/blocks/answer_summary.json
+++ b/schemas/blocks/answer_summary.json
@@ -15,6 +15,9 @@
       "title": {
         "type": "string"
       },
+      "label": {
+        "type": "string"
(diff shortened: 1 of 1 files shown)


MebinAbraham:
From the JSON schema perspective, it only ensures the key `title` is there. If `title` and `label` also needs to be a non-empty string / not none then we need a further check.

jonnyshaw89:
@MebinAbraham isn't that the case for almost every other string?

MebinAbraham:
Is it but its whether that's enough. But can't see us ever writing code to manually check a valid string.

PR #92, opened 2018-10-01
Make optional fields in metadata possible

dcdarrell9:
The lms schema has optional metadata fields for address playback (address_line1, address_line2, etc), which means we need to be able to have extra fields which are optional.

- Test with [this branch in survey-runner](https://github.com/ONSdigital/eq-survey-runner/pull/1804)

- Also running off [the go launcher pr](https://github.com/ONSdigital/go-launch-a-survey/pull/118)

### DIFF
--- a/schemas/questionnaire_v1.json
+++ b/schemas/questionnaire_v1.json
@@ -74,9 +74,27 @@
               "string",
               "uuid"
             ]
+          },
+          "address_line1": {
+            "type": "string"
+          },
+          "address_line2": {
+            "type": "string"
+          },
+          "locality": {
+            "type": "string"
+          },
+          "town_name": {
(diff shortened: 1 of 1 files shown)


jonnyshaw89:
@dcdarrell9 Why?

dcdarrell9:
@jonnyshaw89 Updated pr description

schemas/questionnaire_v1.json
```diff
@@ -76,7 +76,7 @@
             ]
           }
         },
-        "additionalProperties": false,
+        "additionalProperties": true,
```
jonnyshaw89: Instead of blanket allowing extra fields we don't know about. You should define the extra fields you are expecting.

PR #95, opened 2018-10-17
Update validator to handle new answer_count routing

insacuri:
https://trello.com/c/sBtFoRAQ/2482-add-routing-on-the-count-of-multiple-answers-to-when-rule-s

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -19,7 +19,7 @@ class Validator:
         with open('schemas/questionnaire_v1.json', encoding='utf8') as schema_data:
             self.schema = load(schema_data)
 
-    def validate_schema(self, json_to_validate):  # noqa: C901  pylint: disable=too-complex, too-many-branches
+    def validate_schema(self, json_to_validate):  # noqa: C901  pylint: disable=too-complex
 
         schema_errors = self._validate_json_against_schema(json_to_validate)
 
@@ -29,6 +29,7 @@ class Validator:
         errors = []
 
         errors.extend(self._validate_schema_contain_metadata(json_to_validate))
+        errors.extend(self.validate_duplicates(json_to_validate))
 
         numeric_answer_ranges = {}
         answers_with_parent_ids = self._get_answers_with_parent_ids(json_to_validate)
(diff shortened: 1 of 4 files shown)


insacuri:
@MebinAbraham Added duplicate check

schemas/common_definitions.json
```diff
@@ -193,6 +201,11 @@
           "required": [
             "answer_count"
```
MebinAbraham: Pretty sure this is wrong?. This is valid if `oneOf` them is provided but we need more than 1.

app/validation/validator.py
```diff
@@ -302,25 +302,29 @@ def validate_when_rule(when_clause, answer_ids_with_group_id, referenced_id):
         Validates any answer id in a when clause exists within the schema
         Will also check that comparison_id exists
         """
-        answer_reference_id_fields = ('id', 'answer_count', 'comparison_id')
+        answer_reference_id_fields = ('id', 'answer_ids', 'answer_count', 'comparison_id')
```
ajmaddaford: Isn't `answer_count` no longer a property that can exist in a when rule? (as it's now a value of a `type` property)

schemas/common_definitions.json
```diff
@@ -118,6 +118,14 @@
           "type": "string",
           "description": "A repeated group id, for which the count will be fetched."
         },
+        "type": {
+          "type": "string",
+          "description": "Type of rule used for when condition."
```
ajmaddaford: This should be an enum containing "answer_count"

schemas/common_definitions.json
```diff
@@ -193,6 +201,11 @@
           "required": [
             "answer_count"
           ]
+        },
+        {
+          "required": [
+            "type"
```
ajmaddaford: `type` can't be required as most when rules don't define a `type`

app/validation/validator.py
```diff
@@ -302,25 +302,29 @@ def validate_when_rule(when_clause, answer_ids_with_group_id, referenced_id):
...
+                    if when[id_ref_key] not in answer_ids_with_group_id:
+                        return Validator._error_message('The answer id - {} in the {} key of the "when" clause for {} does not exist'
+                                                        .format(when[id_ref_key], id_ref_key, referenced_id))
 
-                if when[id_ref_key] not in answer_ids_with_group_id:
-                    return Validator._error_message('The answer id - {} in the {} key of the "when" clause for {} does not exist'
-                                                    .format(when[id_ref_key], id_ref_key, referenced_id))
-
-                if id_ref_key == 'answer_count':
+                if id_ref_key == 'answer_count' or when.get('type') == 'answer_count':
```
ajmaddaford: This should be moved out of the loop - it doesn't make sense in the loop of `present_id_ref_keys` as the when type does not contain an answer id.

app/validation/validator.py
```diff
@@ -333,6 +337,9 @@ def validate_when_rule(when_clause, answer_ids_with_group_id, referenced_id):
                     return Validator._error_message('The "when" clause for {} contains a comparison_id and uses a '
                                                     'condition of unset or set'
                                                     .format(referenced_id))
+            if 'answer_ids' in when:
+                if len(when['answer_ids']) != len(set(when['answer_ids'])):
+                    return Validator._error_message('Duplicate answer ids found within {} clause'.format(referenced_id))
```
ajmaddaford: I realise this is existing, but I can't see why it returns once it's found one error. The other validation methods build an array of errors and return the array.

schemas/common_definitions.json
insacuri: Isn't this oneOf for the whole block?  I thought it was saying 'you need one of id, meta, answer_count, type', is that wrong?

schemas/common_definitions.json
insacuri: Done

schemas/common_definitions.json
ajmaddaford: My bad, that'll teach me for commenting on github without viewing the whole file 🤦‍♂️

app/validation/validator.py
insacuri: Fixed

insacuri: Fixed

PR #98, opened 2018-10-24
Refactor code to remove ignore pylint error comment

insacuri:
https://trello.com/c/3CGwoO1R/2514-remove-ignore-pylint-comments-in-eq-schema-validator

Quick refactor to remove pylint error.
Also added a few more docstrings that pylint was complaining about.  Nothing too informational but it gives hints as to what it returns which can be useful.

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -19,8 +19,12 @@ class Validator:
         with open('schemas/questionnaire_v1.json', encoding='utf8') as schema_data:
             self.schema = load(schema_data)
 
-    def validate_schema(self, json_to_validate):  # noqa: C901  pylint: disable=too-complex
-
+    def validate_schema(self, json_to_validate):
+        """
+        Validates the json schema provided is correct
+        :param json_to_validate: json schema to be validated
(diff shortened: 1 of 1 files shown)


app/validation/validator.py
```diff
@@ -19,8 +19,12 @@ def __init__(self):
         with open('schemas/questionnaire_v1.json', encoding='utf8') as schema_data:
             self.schema = load(schema_data)
 
-    def validate_schema(self, json_to_validate):  # noqa: C901  pylint: disable=too-complex
-
+    def validate_schema(self, json_to_validate):
```
bitdivision: How about: 
```suggestion
    def validate_schema(self, json_to_validate: str) -> List[Dict[str, str]]:
```

insacuri: I agree.... except a change like this I'd want to do over multiple functions in a single PR.
Would a raised issue/trello card be a fair compromise on it?

bitdivision: Fair enough. I was really just testing suggested changes :)

MebinAbraham:
@insacuri Needs squashing.

PR #99, opened 2018-10-25
Validate default routing rules

bitdivision:
Ensure we always have a default routing rule.

This requires some changes to the schemas in runner to ensure everything validates.

https://github.com/ONSdigital/eq-survey-runner/pull/1863

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -57,9 +57,14 @@ class Validator:
     def _validate_routing_rules(self, group, all_groups, answers_with_parent_ids):
         errors = []
 
+        errors.extend(self.validate_routing_rules_default(group.get('routing_rules', []), group))
         for rule in group.get('routing_rules', []):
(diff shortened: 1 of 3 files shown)


app/validation/validator.py
```diff
@@ -185,6 +188,34 @@ def validate_schema_routing_rule_routes_to_valid_target(self, dict_list, goto_ke
...
+    def validate_routing_rules_default(rules, block_or_group):
+        """
+        Ensure that a set of routing rules contains a default, without a when clause.
+        """
+        errors = []
+
+        if not rules:
+            return errors
+
+        if all(('goto' not in rule for rule in rules)):
```
MebinAbraham: Can we combine these?. Doesn't really affect readability personally.
```
if not rules or all(('goto' not in rule for rule in rules)):
	return errors
```

jonnyshaw89:
@bitdivision is author aware of this change?

PR #104, opened 2019-01-21
Use alpine docker base image

jonnyshaw89:
🤷‍♂️ 
This makes the docker container 52MB down from 377MB


MebinAbraham:
@jonnyshaw89 Needs rebasing.

PR #106, opened 2019-01-23
 Define piping (string interpolation) in a structured way

ajmaddaford:
The current way we define piping (string interpolation) in a schema is to include jinja filters directly in the source strings, which has a number of issues. This pull request contains a proposal to move away from this to a defined schema structure.

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -93,6 +93,8 @@ class Validator:    # pylint: disable=too-many-public-methods
 
             errors.extend(self._validate_questions(block, answers_with_parent_ids, numeric_answer_ranges))
 
+            errors.extend(self._validate_placeholders(block))
+
         return errors
 
     def _validate_questions(self, block, answers_with_parent_ids, numeric_answer_ranges):
@@ -580,6 +582,10 @@ class Validator:    # pylint: disable=too-many-public-methods
 
         for option in answer.get('options', []):
 
+            # labels can have placeholders in, in which case we won't know if they are a duplicate or not
+            if isinstance(option['label'], dict):
+                continue
+
             if option['label'] in labels:
                 errors.append(self._error_message('Duplicate label found - {}'.format(option['label'])))
 
@@ -751,6 +757,71 @@ class Validator:    # pylint: disable=too-many-public-methods
 
         if 'calculated' in answer and ('decimal_places' not in answer or answer['decimal_places'] != 2):
             errors.append(self._error_message("'decimal_places' must be defined and set to 2 for the answer_id - {}".format(answer['id'])))
+
+        return errors
+
+    def _get_dicts_with_key(self, input_data, key_name):
+        """
+        Get all dicts that contain `key_name`.
+        :param input_data: the input data to search
+        :param key_name: the key to find
+        :return: list of dicts containing the key name, otherwise returns None
+        """
+        if isinstance(input_data, dict):
+            for k, v in input_data.items():
+                if k == key_name:
+                    yield input_data
+                else:
+                    yield from self._get_dicts_with_key(v, key_name)
+        elif isinstance(input_data, list):
+            for item in input_data:
+                yield from self._get_dicts_with_key(item, key_name)
+
+    def _validate_placeholders(self, block_json):
+        errors = []
+        strings_with_placeholders = self._get_dicts_with_key(block_json, 'placeholders')
+        for string_with_placeholders in strings_with_placeholders or []:
+            regex = re.compile('{(.*?)}')
+            placeholders_in_string = regex.findall(string_with_placeholders.get('text'))
+            placeholder_definition_names = []
+            for placeholder_definition in string_with_placeholders.get('placeholders'):
+                placeholder_definition_names.append(placeholder_definition['placeholder'])
+
+                transforms = placeholder_definition.get('transforms')
+                if transforms:
+                    errors.extend(self._validate_placeholder_transforms(transforms, block_json['id']))
+
+            if sorted(placeholders_in_string) != sorted(placeholder_definition_names):
+                errors.append(self._error_message(
+                    "Placeholders in 'text' doesn't match 'placeholders' definition for block id '{}'".format(
+                        block_json['id'])))
+
+        return errors
+
+    def _validate_placeholder_transforms(self, transforms, block_id):
+        errors = []
+
+        # First transform can't reference a previous transform
+        first_transform = transforms[0]
+        for argument_name in first_transform.get('arguments'):
+            argument = first_transform['arguments'][argument_name]
+            if isinstance(argument, dict) and argument.get('source') == 'previous_transform':
+                errors.append(self._error_message(
+                    "Can't reference `previous_transform` in a first transform in block id '{}'".format(
+                        block_id)))
+
+        # Previous transform must be referenced in all subsequent transforms
+        for transform in transforms[1:]:
+            previous_transform_used = False
+            for argument_name in transform.get('arguments'):
+                argument = transform['arguments'][argument_name]
+                if isinstance(argument, dict) and argument.get('source') == 'previous_transform':
+                    previous_transform_used = True
+
+            if not previous_transform_used:
+                errors.append(self._error_message(
+                    "`previous_transform` not referenced in chained transform in block id '{}'".format(
+                        block_id)))
 
         return errors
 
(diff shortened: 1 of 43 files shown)


doc/decisions/0001-define-piping-in-a-structured-way.md
```diff
@@ -0,0 +1,310 @@
...
+---
+
+
+## Other notes
+
+- No provision has been made in the schema design for group instances. It is assumed that:
+  - When a placeholder is used in a non-repeating group that any reference to answers resolves to all answers that match the answer id.
+  - When a placeholder is used in a repeating group that any reference to answers resolves to the answer that matches within the current repeat.
+- `format_date_range` and `format_datetime` should be done via two separate placeholders rather than a transform, so that the 'to' and 'at' are in the sentence to be translated.
+- Jinja filters are used to resolve values in the variables schema e.g. `period` in 1_0005. This will no longer work and will require some further design outside of the scope of this proposal.
```
samiwel: I think all of these examples look good. The structured piping approach you've taken is similar to how we've discussed modelling piping in author so we may revisit our implementation at some point and borrow some ideas from here. 👍

There's one scenario that we have a requirement for that I don't see an example of here. Wondering if this would just be another type of transform with two args?

`{{ one_of(metadata['trad_as'], metadata['ru_name']) }}`

ajmaddaford: That's good to hear 🙂

For the `one_of` scenario there is a `first_non_empty_item` filter in runner that takes a list of values and returns the first non-empty one. In new structured json it could look something like this:

```json
"transforms": [
    {
        "transform": "first_non_empty_item",
        "arguments": {
            "items": {
                "source": "metadata",
                "identifier": ["ref_p_start_date","ru_name"]
            },
        }
    }
]
```

or (this would allow for each item to come from a different source):

```json
"transforms": [
    {
        "transform": "first_non_empty_item",
        "arguments": {
            "items": [
				{
                	"source": "metadata",
	                "identifier": "ref_p_start_date"
				},
				{
                	"source": "metadata",
	                "identifier": "ru_name"
				}
            ]
        }
    }
]
```

samiwel: Nice, yeah that makes sense

doc/decisions/0001-define-piping-in-a-structured-way.md
```diff
@@ -0,0 +1,310 @@
...
+
+---
+
+
+## Other notes
+
+- No provision has been made in the schema design for group instances. It is assumed that:
+  - When a placeholder is used in a non-repeating group that any reference to answers resolves to all answers that match the answer id.
+  - When a placeholder is used in a repeating group that any reference to answers resolves to the answer that matches within the current repeat.
+- `format_date_range` and `format_datetime` should be done via two separate placeholders rather than a transform, so that the 'to' and 'at' are in the sentence to be translated.
```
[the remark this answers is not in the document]
ajmaddaford: Piping is not currently supported in any element of a content block. I think it's omission from the schema is documentation enough that this isn't supported. I've been careful to only add piping to elements that currently make use of piping, so it's clear where we use it. If we start using piping in content blocks we can update the schema to reflect that.

schemas/string_interpolation/definitions.json
```diff
@@ -0,0 +1,143 @@
...
+    },
+    "minItems": 1
+  },
+  "placeholder": {
+    "description": "A mapping of a string placeholder to it's value. The value can also be transformed.",
+    "type": "object",
+    "properties": {
+      "placeholder": {
+        "description": "The placeholder in the string that will be replaced e.g. 'test' is the placeholder in 'My {test} string'.",
+        "type": "string"
```
MebinAbraham: `placeholder` within `placeholders` should be unique. Don't see a reason for it to not be?.

schemas/string_interpolation/definitions.json
```diff
@@ -0,0 +1,143 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "string_with_placeholders": {
+    "oneOf": [
```
MebinAbraham: We should have additional properties false at this level to avoid odd behaviour when using string type.
```json
"string_with_placeholders": {
	"additionalProperties": false,
	"oneOf": [
		...
```

schemas/string_interpolation/definitions.json
```diff
@@ -0,0 +1,143 @@
...
+          "text": {
+            "type": "string"
+          },
+          "placeholders": {
+            "$ref": "#/placeholders"
+          },
+          "additionalProperties": false,
+          "required": [
+            "text",
+            "placeholders"
```
MebinAbraham: These don't seem to be enforced for me. If I define:
```
"title": {
	"text": "hello"
}
This passes validator for me.

File: schemas/string_interpolation/definitions.json (hunk 3 above)
Comment (ajmaddaford) [reply schemas/string_interpolation/definitions.json]: Yes, I'll look at adding some validation for that. I'll also look at adding validation to check that the placeholders in the source string are defined in the placeholders definition.

File: schemas/string_interpolation/definitions.json (hunk 5 above)
Comment (ajmaddaford) [reply schemas/string_interpolation/definitions.json]: Good spot! `additionalProperties` and `required` were at the wrong level in the schema. Fixed.

File: schemas/string_interpolation/definitions.json [hunk 6]
```diff
@@ -0,0 +1,143 @@
...
+      {
+        "$ref": "#/chained_source"
+      },
+      {
+        "$ref": "#/lookup_source"
+      }
+    ]
+  },
+  "chained_source": {
+    "description": "This will take the value from the previous transform. Can't be used for the first transform.",
```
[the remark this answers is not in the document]
MebinAbraham: @ajmaddaford Unless I'm mistaken, if the length of `transforms` for a given placeholder is > 1 then we should enforce the `source` for all but the first transform must be `previous_transform`?

ajmaddaford:
@MebinAbraham @bitdivision comments addressed, ready for re-review.

schemas/string_interpolation/definitions.json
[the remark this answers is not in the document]
ajmaddaford: Makes sense. I've added validation for:
- `previous_transform` can't be used in a first transform
- Subsequent transforms must reference `previous_transform` for at least one argument

app/validation/validator.py
```diff
@@ -754,6 +756,73 @@ def validate_totaliser_defines_decimal_places(self, answer):
...
+            for placeholder_definition in string_with_placeholders.get('placeholders'):
+                placeholder_definition_names.append(placeholder_definition['placeholder'])
+
+            if sorted(placeholders_in_string) != sorted(placeholder_definition_names):
+                errors.append(self._error_message(
+                    "Placeholders in 'text' doesn't match 'placeholders' definition for block id '{}'".format(
+                        block_json['id'])))
+
+            for placeholder_definition in string_with_placeholders.get('placeholders'):
+                transforms = placeholder_definition.get('transforms')
```
MebinAbraham: Check both scenarios in the first and one loop?

MebinAbraham:
@ajmaddaford We should allow `string_with_placeholders` for `description` for the different question types. It's used by Census. Its only enabled for titles at the moment.

ajmaddaford:
@MebinAbraham comments addressed, ready for re-review 👀

schemas/answers/definitions.json
```diff
@@ -11,7 +11,7 @@
       "type": "object",
       "properties": {
         "label": {
-          "type": "string"
+          "$ref": "../string_interpolation/definitions.json#/string_with_placeholders"
```
MebinAbraham: Not able to use `string_with_placeholders` on option labels because it errors on existing check that doesn't know about the new possible structure of label. So needs some more code changes sry 👀 
```
 File "/Users/abrahm/Documents/projects/eq-schema-validator/app/validation/validator.py", line 115, in _validate_questions
    errors.extend(self.validate_duplicate_options(answer))
  File "/Users/abrahm/Documents/projects/eq-schema-validator/app/validation/validator.py", line 585, in validate_duplicate_options
    if option['label'] in labels:
TypeError: unhashable type: 'dict'
```

Also some minor linting issues:
```
************* Module app.validation.validator
app/validation/validator.py:804:0: C4001: Invalid string quote ", should be ' (invalid-string-quote)
app/validation/validator.py:814:0: C4001: Invalid string quote ", should be ' (invalid-string-quote)
```

ajmaddaford: I've added some test schema for placeholders in radio and checkbox option labels.

MebinAbraham:
@ajmaddaford Squishy Squashy

PR #107, opened 2019-01-24
Change proxy implementation to be more flexible

bitdivision:
The current implementation of proxy has limitations in terms of what it can be used for. This proposes new schema formats for interstitial and question to allow multiple variants per block.

I haven't made any changes to the JSON Schema, but will do so once this has been discussed and approved.

### DIFF
--- a/doc/decisions/0002-define-proxy-mechanism.md
+++ b/doc/decisions/0002-define-proxy-mechanism.md
@@ -1,0 +1,352 @@
+# 2. Define a more flexible approach to proxy questions
+
+## Context
+Currently, the solution for proxy question titles has some limitations:
+
+- Unable to modify any part of a question except the title
+- Current solution difficult to integrate with structured string interpolation.
+- Does not support interstitials
+
+Since only titles are currently changeable, routing and duplicated blocks have been used to resolve this, but this can cause issues when routing becomes complicated or changes need to be made later on.
+
+### Example of current question schema
+
+Currently, proxy titles are defined through the question `titles` object:
+
+ ```json
+{
+    "id": "example",
+    "type": "Question",
+    "questions": [{
+        "id": "example-question",
+        "titles": [{
+                "value": "Did they work today?",
+                "when": [{
+                    "id": "proxy-check-answer",
+                    "condition": "equals",
+                    "value": "proxy"
+                }]
+            },
+            {
+                "value": "Did you do any work today?"
+            }
+        ],
+        "type": "General",
+        "answers": [{
+            "id": "example-answer",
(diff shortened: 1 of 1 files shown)


doc/decisions/0002-define-proxy-mechanism.md
```diff
@@ -0,0 +1,342 @@
...
+                        "Some of these things are needed"
+                    ]
+                },
+                {
+                    "description": "Some more extra information"
+                }
+            ]
+        },
+        "past-main-employment-proxy": {
+            "id": "past-main-employment-proxy",
```
MebinAbraham: Shouldn't be here?

bitdivision:
@MebinAbraham I would expect a single variant with a valid variant_choice to be shown as normal, however the suggested pattern would be to change that single variant to a `question`. See [line 338](https://github.com/ONSdigital/eq-schema-validator/pull/107/files#diff-e6e4c02a7de87002ff67a7de05f7e046R338)

MebinAbraham:
> @MebinAbraham I would expect a single variant with a valid variant_choice to be shown as normal, however the suggested pattern would be to change that single variant to a `question`. See [line 338](https://github.com/ONSdigital/eq-schema-validator/pull/107/files#diff-e6e4c02a7de87002ff67a7de05f7e046R338)

A single variant wouldn't have a variant_choice?. 
But we would no longer support `questions` in the schema is what I mean.

bitdivision:
The behaviour for no variant_choice isn't currently defined. I think you're right though, it would make sense to display a single variant if it's the only one defined (even without variant_choice). I'll add that to the doc

bitdivision:
I have disallowed variants of length 1 since these should be converted to `question` objects.

samiwel:
@bitdivision Ah I see, I guess that makes sense. From an author point of view we only ever allow users to publish schemas that have a single question per block. I guess it makes sense to drop the `questions` array in favour of a single `question` but if it could be implemented in two separate PRs (i.e. keep `questions` around and also add `question` then a separate PR to drop `questions`) then it would allow us to make necessary changes on our end and ensure that the validator change is non-breaking.

MebinAbraham:
Since each `variant` is independent of each other, will things like `type` be normalised?

bitdivision:
@MebinAbraham - Yes, I want to support all of the same features, a variant should be exactly the same as a question object.

I don't expect any serious performance loss, the json will take slightly longer to parse, but we should be able to cache this I think. Don't think there's any need to worry about cache locality etc.

bitdivision:
Just going to change this so that the ID is still inside the question object.

PR #108, opened 2019-01-25
Ensure that answer option values do not contain piping

tgandrews:
Validation to enforce that option values cannot contain piping.

I contemplated creating a whitelist of allowed characters but everything seems to accept string so I presume that would be restrictive. Instead I have done a very simple regex test that checks to see if anything looks like a piping value. However this might not be optimal. Suggestions welcome.

## Testing
1. You can apply a piping value to an option label and publisher will send it as the value so I tested this against publisher using that but you have to change the publisher validator URL to `docker.for.mac.localhost:5001` once you have schema-validator running locally.
2. The unit tests should pass.

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -111,6 +111,7 @@ class Validator:    # pylint: disable=too-many-public-methods
             for answer in question.get('answers', []):
                 errors.extend(self.validate_routing_on_answer_options(block, answer))
                 errors.extend(self.validate_duplicate_options(answer))
+                errors.extend(self.validate_option_values(answer))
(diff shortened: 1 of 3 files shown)


tgandrews:
[UKIS Issue with piping going into an answer value](https://trello.com/c/Wqo5RMKh/845-ukis-issue-with-piping-going-into-an-answer-value)

tgandrews:
@samiwel I have resolved the linting issues.

samiwel:
@dwyeradam Do we need to get someone on the Census team to review?

PR #109, opened 2019-01-30
Use Travis instead of Concourse

MebinAbraham:
Have decided to use Travis to run checks on PRs and merges instead of Concourse to keep it consistent with runner.
Since Concourse is on our infrastructure it also means that any arbitrary code could be run when someone raises a PR.

Ensure:
- Travis CI works as expected. (Runs linting, tests, and pushed tag to Docker Hub)


MebinAbraham:
I will set up automated build on docker up to updates the images on pushes etc.

MebinAbraham:
Waiting on docker hub state to update -_-

.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+dist: trusty
+sudo: true
+before_install:
+    - pyenv versions
+    - pyenv install --skip-existing
+    - pyenv rehash
+    - pyenv global $(cat .python-version)
+    - pip install -U pip pipenv wheel
+    - pipenv install --dev --deploy
+    - pipenv check -i '36810'
```
bitdivision: Can we add a comment saying what this is?

Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM python:3.6-alpine
 
-RUN pip install pipenv==8.2.7
+RUN pip install pipenv==2018.11.26
```
bitdivision: Does this need to be pinned?

MebinAbraham: Codacy suggests that it should be and seems to be a pattern in other repos also.

.travis.yml
MebinAbraham: Added

MebinAbraham:
@bitdivision @jonnyshaw89 
Have made a slight change, instead of ignoring the issue with pre-installed numpy during `pipenv check`, I've set a flag to force Travis to recreate the virtual env.

PR #111, opened 2019-02-07
Ensure that required fields are actually populated

tgandrews:
Required fields such as labels on options and some answer types only had to be a string - their contents was never validated. This meant that empty strings were viewed as being valid. This meant that runner could be made to look broken.

This is a breaking change for some of the test schemas in runner and here is the PR to resolve this https://github.com/ONSdigital/eq-survey-runner/pull/1984

### DIFF
--- a/schemas/common_definitions.json
+++ b/schemas/common_definitions.json
@@ -10,6 +10,10 @@
     "description": "A question code used by downstream systems to identify answers.",
     "pattern": "^[0-9a-z]+$"
   },
+  "populated_string": {
+    "type": "string",
+    "pattern": "\\w+"
+  },
   "definitions": {
     "description": "Allows customisation of question definition title and description.",
     "type": "array",
--- a/schemas/answers/currency.json
+++ b/schemas/answers/currency.json
@@ -10,7 +10,7 @@
         "$ref": "../common_definitions.json#/q_code"
       },
       "label": {
-        "type": "string"
+        "$ref": "../common_definitions.json#/populated_string"
       },
       "guidance": {
         "$ref": "../common_definitions.json#/guidance"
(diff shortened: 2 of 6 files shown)


tgandrews:
[Able to publish when invalid state](https://trello.com/c/naZhrAsC/917-able-to-publish-when-invalid-state)

schemas/common_definitions.json
```diff
@@ -10,6 +10,10 @@
     "description": "A question code used by downstream systems to identify answers.",
     "pattern": "^[0-9a-z]+$"
   },
+  "populated_string": {
+    "type": "string",
+    "pattern": "\\w+"
```
samiwel: Rather than restricting to word values, should be try and match a non-whitespace character i.e `\\S+`?

samiwel: This would allow us to use non-word chars in options - perhaps we would never do it, but it was possible to do that in the past.

MebinAbraham: This wouldn't work with the schemas currently in runner.

MebinAbraham:
Curious, why have only a few of these changed?.

schemas/common_definitions.json
MebinAbraham: @samiwel We do need to allow non-word characters ie apostrophes. However, this only ensures that there is at least 1 word character in the string to be classed as valid. So a string with non-word character with atleast 1 word character would pass.

samiwel: Yes that's correct, Mebin. Labels containing apostrophes and other punctuation should be fine so long as they also contain at least one word character.

The more I think about it, I think this pattern is fine for now. Just because it's valid to have a label with a single character in it, doesn't mean we (the EQ team) would allow such a schema to be made live. It's more of a safety net to catch cases where the label may have accidentally not been specified.

Tom, ignore my previous comment regarding the pattern happy to give my approval to this change.

PR #113, opened 2019-02-15
Define new variant schema

MebinAbraham:
Card: https://trello.com/c/3r1PVjpm/2753-update-json-schema-and-validator-for-new-variant-schema
This change implements the rules and restrictions of the new variant schema.
- Allows questions to be defined as `questions` or `variants`.

### How to review

- Ensure changes are as per [spec](https://github.com/ONSdigital/eq-schema-validator/blob/bb7e852fb81ae73cfc7d311284a2ec3b14df741d/doc/decisions/0002-define-proxy-mechanism.md).

- Ensure variants can be used by types like `Interstitials`, `Confirmation` as well as `Question` types.

Was unsure if the naming should be changed to `question_variants` and `content_variants`.


app/validation/validator.py
```diff
@@ -86,21 +86,57 @@ def _validate_blocks(self, json_to_validate, section, group, all_groups, answers
...
+        total_choices_without_whens = len([c for c in variant_choices if not c.get('when')])
+        if total_choices_without_whens > 1:
+            errors.append(self._error_message('Multiple defaults found in `variant_choice`'))
+
+        return errors
+
     def _validate_questions(self, block, answers_with_parent_ids, numeric_answer_ranges):
         errors = []
 
-        for question in block.get('questions', []):
```
bitdivision: Do we need a validate_content() too?

schemas/blocks/confirmation_question.json
```diff
@@ -40,6 +45,27 @@
       "id",
       "type",
       "questions"
+    ],
+    "oneOf": [
```
jonnyshaw89: Why does this not apply to all blocks?

bitdivision: Because interstitial doesn't have `questions`?

bitdivision: Ah, it does

jonnyshaw89: We should define this as a common type and reuse the same definition

bitdivision: This is seemingly difficult to extract with current versions of json schema.

app/validation/validator.py
```diff
@@ -881,14 +925,12 @@ def _get_answers_with_parent_ids(json_to_validate):
...
-                        for question in block['questions']:
-                            for answer in question.get('answers', []):
-                                answers[answer['id']] = {
-                                    'answer': answer,
-                                    'block': block['id'],
-                                    'group_id': group['id'],
-                                    'section': section['id']
-                                }
-
+                    for question_or_variant in block.get('questions', []) + list(block.get('variants', {}).values()):
```
jonnyshaw89:
```suggestion
                    for question_or_variant in block.get('questions', list(block.get('variants', {}).values())):
```

app/validation/validator.py
```diff
@@ -48,7 +48,7 @@ def validate_schema(self, json_to_validate):
                 errors.extend(self._validate_routing_rules(group, all_groups, answers_with_parent_ids))
 
                 for skip_condition in group.get('skip_conditions', []):
-                    errors.extend(self.validate_skip_condition(skip_condition, answers_with_parent_ids, group))
+                    errors.extend(self.validate_display_condition(skip_condition, answers_with_parent_ids, group))
```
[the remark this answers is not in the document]
bitdivision: Reverted.

app/validation/validator.py
```diff
@@ -86,21 +86,57 @@ def _validate_blocks(self, json_to_validate, section, group, all_groups, answers
...
+    def _validate_variants(self, block):
+        errors = []
+
+        variants = [v for v in block.get('variants', {}).keys()]
+        variant_choices = block.get('variant_choice', [])
+
+        if variant_choices and not variants:
+            errors.append(self._error_message(f"Missing `variants` in block `{block['id']}`. Cannot define `variant_choice` without `variants`"))
+        elif variants and not variant_choices:
+            errors.append(self._error_message(f"Missing `variant_choice` in block `{block['id']}`"))
```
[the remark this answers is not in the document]

app/validation/validator.py
```diff
@@ -86,21 +86,57 @@ def _validate_blocks(self, json_to_validate, section, group, all_groups, answers
...
             errors.extend(self._validate_questions(block, answers_with_parent_ids, numeric_answer_ranges))
 
             errors.extend(self._validate_placeholders(block))
 
         return errors
 
+    def _validate_variants(self, block):
+        errors = []
+
+        variants = [v for v in block.get('variants', {}).keys()]
```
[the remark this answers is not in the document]

app/validation/validator.py
```diff
@@ -86,21 +86,57 @@ def _validate_blocks(self, json_to_validate, section, group, all_groups, answers
...
+            if choice['variant'] not in variants:
+                errors.append(self._error_message(f"Variant `{choice['variant']}` not found in `variants`"))
+
+        for variant in variants:
+            if variant not in [c['variant'] for c in variant_choices]:
+                errors.append(self._error_message(f'Variant `{variant}` not found in `variant_choice`'))
+
+        total_choices_without_whens = len([c for c in variant_choices if not c.get('when')])
+        if total_choices_without_whens > 1:
+            errors.append(self._error_message('Multiple defaults found in `variant_choice`'))
```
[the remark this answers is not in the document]

app/validation/validator.py
bitdivision: Fixed.

ajmaddaford:
I think we should consider reworking the schema to enforce more of what we've ended up having to validate in code. I've spent some time talking this through with @jonnyshaw89 and we've come up with an alternative:

```json
"variants": [
    { 
        "question": {
            "id": "question-2",
            "type": "General",
            "title": "What is your age?",
            "answers": [{
                "id": "answer-2",
                "label": "Your age?",
                "mandatory": false,
                "type": "Number"
            }]
        },
        "when": [{
            "id": "answer-1",
            "condition": "greater than",
            "value": "16"
        }]
   }
]
```

This means that:
- We can define a question variant schema type that brings together a question and a when rule. 
- We don't need to have another identifier to link the when rule to the variant.
- We can remove the code to validate matching variants to when rules.

Aside from that:

- I think the `when` rule should be enforced for all variants as we don't have any examples where a default makes sense. We should be able to validate that at least one `when` rule would be selected when validating a schema.

- We should allow duplicate answer ids in question variants, as one of the reasons for doing this is to simplify any subsequent routing - so we can route based on one answer id rather than all it's variants.

app/validation/validator.py
```diff
@@ -578,18 +579,30 @@ def _get_numeric_range_values(self, answer, answer_ranges):
...
-                if value in unique_items:
-                    duplicate_errors.append(
-                        self._error_message('Duplicate {} found. value {}'.format(special_key, value)))
-                else:
-                    unique_items.append(value)
+            items_per_block = defaultdict(set)
+
+            for path, value in self._parse_values(json_to_validate, special_key):
+                if 'blocks' in path:
+                    try:
```
iwootten: Is there a better way of handling this? Check for key existence prior to insertion?

bitdivision: Potentially, I know exceptions aren't liked much around here. Looking at it, this should also be catching IndexError in case the path ends early, so the check becomes a bit more complicated.

bitdivision: Actually, this only needs indexerror, and that can be handled more simply.

bitdivision:
I have also added the ability to use `question` rather than `questions`. `questions` can be removed once the schema changes have been made in runner.

bitdivision:
> If I take the example given in the PR and flesh it out as a survey [gist.github.com/iwootten/a2a1e5ee406e0f395ec83003a6b3e18c](https://gist.github.com/iwootten/a2a1e5ee406e0f395ec83003a6b3e18c) - get the following when I attempt to validate:
> 
> ```
>     "path": "deque(['sections', 0, 'groups', 0, 'blocks', 0])",
>     "predicted_cause": "Additional properties are not allowed ('question' was unexpected)"
> ```

This is actually because the variant only has a single question in it, and variants defines the minItems on variants to be two. I agree that the error message is awful, I can potentially cover this case in the python just to give a better error?


> Additionally, does it make sense to update the documentation to match Andrews comments?

I'll update this.

app/validation/validator.py
```diff
@@ -95,12 +96,20 @@ def _validate_blocks(self, json_to_validate, section, group, all_groups, answers
 
             errors.extend(self._validate_placeholders(block))
 
+            errors.extend(self._validate_variants(block, answers_with_parent_ids, numeric_answer_ranges))
+
         return errors
 
     def _validate_questions(self, block, answers_with_parent_ids, numeric_answer_ranges):
```
MebinAbraham: In this content, is block still being referred to as our schema block or just a JSON block because technically its `block_or_variants`

app/validation/validator.py
```diff
@@ -881,14 +922,36 @@ def _get_answers_with_parent_ids(json_to_validate):
...
-                        for question in block['questions']:
-                            for answer in question.get('answers', []):
-                                answers[answer['id']] = {
-                                    'answer': answer,
-                                    'block': block['id'],
-                                    'group_id': group['id'],
-                                    'section': section['id']
-                                }
+                    for variant in block.get('variants', []):
+                        for answer in variant.get('question', {}).get('answers', []):
```
MebinAbraham: Is it worth combining variants, questions and question in one place to avoid some dupe below, as long as it's still readable? 🤷‍♂️

schemas/blocks/interstitial.json
```diff
@@ -23,15 +23,13 @@
...
-            }
-          ]
-        }
+        "$ref": "../questions/definitions.json#/content_questions"
+      },
+      "question": {
+        "$ref": "../questions/definitions.json#/question"
+      },
+      "variants": {
+        "$ref": "../questions/definitions.json#/variants"
```
MebinAbraham: Should this not be `content_variants`?

schemas/blocks/interstitial.json
```diff
@@ -23,15 +23,13 @@
...
-          "oneOf": [
-            {
-              "$ref": "../questions/content.json#/question"
-            }
-          ]
-        }
+        "$ref": "../questions/definitions.json#/content_questions"
+      },
+      "question": {
+        "$ref": "../questions/definitions.json#/question"
```
MebinAbraham: Need a `content_question`?

schemas/blocks/confirmation.json
```diff
@@ -19,13 +19,19 @@
         "type": "string",
         "enum": ["Confirmation"]
       },
+      "question": {
+        "$ref": "../questions/types/content.json#/question"
```
MebinAbraham: We need a `content_question`?

schemas/questions/definitions.json
```diff
@@ -0,0 +1,52 @@
...
+    "minItems": 2,
+    "items": {
+      "$ref": "variants/content_variant.json#/variant"
+    }
+  },
+  "question": {
+    "oneOf": [{
+      "$ref": "types/general.json#/question"
+    },
+      {
```
MebinAbraham: Formatting 👀

schemas/questions/types/content/content.json
```diff
@@ -0,0 +1,53 @@
+{
```
MebinAbraham: This file is essentially a duplicate of `schemas/questions/types/content.json`.

schemas/questions/types/content/definitions.json
```diff
@@ -0,0 +1,24 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "questions": {
+    "type": "array",
+    "minItems": 1,
+    "items": {
+      "$ref": "content.json#/question"
+    }
+  },
+  "variants": {
+    "patternProperties": {
```
MebinAbraham: This is redundant since Andrew's input?

app/validation/validator.py
bitdivision: I'm not sure if this comment is on the wrong line? _validate_questions takes a block in terms of our schema block.

app/validation/validator.py
bitdivision: I flip flopped on this one. It's going to be changed fairly soon when `questions` is removed. But yeah, can probably be simpler. I'll change it.

schemas/blocks/interstitial.json
bitdivision: I didn't want to remove backwards compatibility for question on interstitial at this point.

schemas/blocks/interstitial.json
bitdivision: As above.

schemas/questions/types/content/content.json
bitdivision: Sorry, I had meant to delete that folder

app/validation/validator.py
MebinAbraham: As discussed, referring to the fact that `_validate_questions` are both called on its own with a `block` and inside `_validate_variants` but with variants.

bitdivision: :+1: Yup, I misunderstood.

bitdivision:
I think I've addressed all the comments @MebinAbraham 	@iwootten 	

Only question left is whether to enforce min length of variants via python rather than json schema.

iwootten:
I'd prefer helpful error messages as a dev, so I don't have to bother anyone else. Doe we need to look at what library we're using to produce these messages and use something more verbose?

bitdivision:
Agree that the json schema library should be changed. 'ajv' is probably the best bet here, I think @ajmaddaford 	started work on this.

I don't want to swap to ajv as part of this PR, since it's likely to be a fairly large change. 

For now, I think adding an error to the python validator is probably the best option. The annoying thing is that the schema validation is run before our python validator runs, but that can be changed to not return immediately afterwards.

app/validation/validator.py
```diff
@@ -881,14 +922,35 @@ def _get_answers_with_parent_ids(json_to_validate):
...
+                                'group_id': group['id'],
+                                'section': section['id']
+                            }
 
         return answers
+
+
+def find_duplicates(lists):
+    seen = set()
+    for lst in lists:
```
jonnyshaw89: 😢 `list` not `lst`

app/validation/validator.py
```diff
@@ -559,18 +579,27 @@ def _get_numeric_range_values(self, answer, answer_ranges):
...
-                if value in unique_items:
-                    duplicate_errors.append(
-                        self._error_message('Duplicate {} found. value {}'.format(special_key, value)))
-                else:
-                    unique_items.append(value)
+            for path, value in self._parse_values(json_to_validate, special_key):
+                if 'blocks' in path:
+                    block_path = path[:path.index('blocks') + 2]
+                    string_path = '.'.join(block_path)
+                    items_per_block[string_path].add(value)
```
jonnyshaw89: Maybe worth a comment around this bit of code, took me a while to work out how it was working

app/validation/validator.py
bitdivision: list is a builtin. I'll think of something else.

app/validation/validator.py
bitdivision: Agreed, it's not immediately obvious. Maybe there's a simpler way to do it?

jonnyshaw89: I like the approach, using the uniqueness of a set, just maybe a comment to highlight thats what it is doing

bitdivision:
@MebinAbraham 	@jonnyshaw89 	@ajmaddaford 	@iwootten 
I've made some non functional changes to the tests to stop using unittest and cleanup some bits. 

At this point should be fine to review though.

bitdivision:
Formatting is not enforced anywhere at the moment. I can add another gulp task to verify formatting with a linter though ala runner.

bitdivision:
I've fixed all of the test schemas (to the point that the tests pass). Going through them though, it's clear that most of them contain a lot of functionality that is completely extraneous to the feature being tested. This could probably do with being cleaned up at some point (e.g. one egregious example was testing that whitespace wasn't allowed in a survey id by using the LMS alpha schema).

Since we are now breaking runner schemas, it's probably also worth removing the current title functionality and converting the test schemas as part of this PR too.

bitdivision:
I have removed all reference to titles from the schemas now. The test schemas have been switched to variants where needed.

I think the idea now is to leave this PR open (as long as everyone is happy with it), and start work on the runner implementation against this branch of validator.

bitdivision:
@ajmaddaford  suggested that we might want to change `variants` to `question_variants` and `content_variants`. (see https://github.com/ONSdigital/eq-survey-runner/pull/2011#discussion_r262598394)

For some reason I like having a top-level generic key. Maybe because specialising adds another level of duplication and more block level keys? e.g.

```
"question_variants": [
    {  
        "question": {          
            "id": "question-id",       
            ...
        },  
        "when": {
            ...
        }    
    }, 
    {
        "question": {
            "id": "another-question-id",
            ...
        },
        "when": {
            ...
        }
    } 
]
```

Realistically though, the objects are specialised (we never want both content and question in the same variant) and having it at the top level might make runner implementation slightly simpler.

I'm happy to make that change unless anyone has any objections?

jonnyshaw89:
@ajmaddaford @bitdivision im happy with `variants`

ajmaddaford:
My original reasoning (in the other pull request) was when there was the potential to have things like `add_question_variants` at the top level as that would mean there wouldn’t be an equivalent association of question to it’s variants. Given we’re going to move that into blocks that's not an issue. 

However, I think we should change to `question_variants` for a clearer association with the `question` element.

bitdivision:
@ajmaddaford @jonnyshaw89 I've made the change to `question_variants` and `content_variants`.

doc/decisions/0002-define-proxy-mechanism.md
```diff
@@ -264,13 +269,11 @@ To allow proxy versions of interstitials, a top-level `variants` object has been
 }
 ```
 
-Variant choice is composed of a list of objects, these will be evaluated in the order of the list and the first matching variant will be chosen for display.
+Each variant object can contain a content object and a when rule which governs whether the question / content should be displayed.
```
ajmaddaford: question is not relevant for content blocks

doc/decisions/0002-define-proxy-mechanism.md
```diff
@@ -319,34 +329,25 @@ For questions, the basic premise is to use the multiple question scheme that is
...
 }
 ```
 
-Each variant should have the same structure as questions currently do, including the `id`. The key of each object in `variants` is a label for that variant and does not necessarily correspond to the question id.
-
-Skip conditions on questions have been moved to the top level of the block and into `variant_choice`. This enforces a seperation between the questions themselves and the logic around whether or not they should be displayed.
+Each variant should have the same structure as questions currently do. 
 
-The schema should intitially still allow `question` to be used for questions which don't require a variant form. The question should still use the same object schema as currently found within `questions`.
+The schema should initially still allow `questions` to be used for questions which don't require a variant form. This will eventually be changed to use `question` and enforce that it contains a single question.
```
ajmaddaford: I thought we were removing `questions` as part of the breaking changes we're making in this pr?

app/validation/validator.py
```diff
@@ -833,20 +851,34 @@ def _is_contained_in_list(dict_list, key_id):
 
         return False
 
-    def _parse_values(self, schema_json, parsed_key):
-        ignored_keys = ['titles', 'routing_rules', 'skip_conditions']
+    def _parse_values(self, schema_json, parsed_key, path=None):
+        """ generate a list of values with a key of `parsed_key`.
+
+        These values will be returned with the path to them through the object e.g.
+            - (['sections', '[0]', 'groups', '[0]', 'blocks', '[1]', 'question_variants', '[0]', 'question', 'question-2')
```
ajmaddaford: @iwootten has been using json pointers to keep references to specific elements in the schema for the piping and translation work. It would be worth seeing if it's worth using that here.

bitdivision: Good point, I considered it early on and then didn't revisit it for some reason. Will try now

doc/decisions/0002-define-proxy-mechanism.md
bitdivision: Yep, I just failed to update it.

app/validation/validator.py
bitdivision: I've modified this to return JSON pointers. I haven't used a library here, since https://python-json-pointer.readthedocs.io/en/latest/mod-jsonpointer.html doesn't support building paths like this very well.

app/validation/validator.py
```diff
@@ -559,18 +568,30 @@ def _get_numeric_range_values(self, answer, answer_ranges):
...
+        """
+        question_id & answer_id should be globally unique with some exceptions:
+            - within a block, ids can be duplicated across variants, but must still be unique outside of the block.
+        """
+
         duplicate_errors = []
 
-        for special_key in special_keys:
-            unique_items = []
+        items_per_block = defaultdict(set)
```
ajmaddaford: I think this would be clearer if we named it something like `unique_ids_per_block`

app/validation/validator.py
```diff
@@ -559,18 +568,30 @@ def _get_numeric_range_values(self, answer, answer_ranges):
...
+                # Generate a string path and add it to the set representing the ids in that path
+                path_list = path.split('/')
+
+                block_path = path_list[:path_list.index('blocks') + 2]
+
+                string_path = '/'.join(block_path)
+                # Using a set for items_per_block allows duplicate ids within a specific block.
+                items_per_block[string_path].add(value)
+
+        duplicates = find_duplicates(items_per_block.values())
```
ajmaddaford: I think this now only checks block level ids and below i.e. it's not checking group ids for uniqueness.

app/validation/validator.py
bitdivision: Fixed

app/validation/validator.py
bitdivision: Fixed

app/validation/validator.py
```diff
@@ -585,10 +586,15 @@ def validate_duplicates(self, json_to_validate):
...
 
                 string_path = '/'.join(block_path)
-                # Using a set for items_per_block allows duplicate ids within a specific block.
-                items_per_block[string_path].add(value)
+                # Using a set for unique_ids_per_block allows duplicate ids within a specific block.
+                unique_ids_per_block[string_path].add(value)
+            else:
+                non_block_ids.append(value)
+
+        all_ids = list(unique_ids_per_block.values()) + [[x] for x in non_block_ids]
```
ajmaddaford: I think this would be simpler if we flattened this down to just a list of all ids before checking for duplicates.

bitdivision:
Hopefully ready to go now once Codacy gets its act together.

PR #114, opened 2019-02-26
Validate that section title exists when navigation is enabled

tgandrews:
This ensures that when navigation is visible that section title must be populated. Without this change when no section title is provided and navigation is visible this results in runner looking broken.


tgandrews:
[Section/survey introduction review (2)](https://trello.com/c/av5w2aOd/800-section-survey-introduction-review-2)

PR #115, opened 2019-02-28
Use non global named args

iwootten:

Updates the arguments to placeholder transforms to be less generic and non conflicting with python globals. This allows transformations to meet our coding standards in runner.

Additionally add format_currency as a transform, as defined in the docs here: https://github.com/ONSdigital/eq-schema-validator/blob/v3/doc/decisions/0001-define-piping-in-a-structured-way.md#previous-answers-or-metadata-transformed-in-some-way

Once this is merged, https://github.com/ONSdigital/eq-survey-runner/pull/2010 will pass validation tests.

### DIFF
--- a/schemas/string_interpolation/definitions.json
+++ b/schemas/string_interpolation/definitions.json
@@ -56,6 +56,9 @@
               "$ref": "transforms/format_number.json#/format_number"
             },
             {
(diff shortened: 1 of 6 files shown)


bitdivision:
@iwootten 	Needs a rebase.

PR #116, opened 2019-03-11
Add lists to schema validator

jonnyshaw89:
Added a `ListCollector` Block type as defined in https://github.com/ONSdigital/eq-survey-runner/blob/v3/doc/architecture/decisions/0006-make-named-lists-a-first-class-construct.md

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -1,10 +1,8 @@
 import os
 import re
-
 import pathlib
 from json import load
 from collections import defaultdict
-
 from datetime import datetime
 from dateutil.relativedelta import relativedelta
 from jsonschema import SchemaError, RefResolver, validate, ValidationError
@@ -100,10 +98,11 @@ class Validator:    # pylint: disable=too-many-public-methods
             if block['type'] == 'CalculatedSummary':
                 errors.extend(self.validate_calculated_summary_type(block, answers_with_parent_ids))
 
+            if block['type'] == 'ListCollector':
+                errors.extend(self.validate_list_collector_type(block))
+
             errors.extend(self._validate_questions(block, numeric_answer_ranges))
-
             errors.extend(self._validate_placeholders(block))
-
             errors.extend(self._validate_variants(block, answers_with_parent_ids, numeric_answer_ranges))
 
         return errors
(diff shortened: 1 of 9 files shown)


app/validation/validator.py
```diff
@@ -312,6 +315,38 @@ def validate_repeat_rule_restricted(self, rule, block):
...
+                questions.append(question)
+
+        for question in questions:
+            for answer in question['answers']:
+                if answer['type'] != 'Radio':
+                    errors.append(self._error_message('The block {} contains a list collector without a Radio answer type'
+                                                      .format(block['id'])))
+
+        if 'routing_rules' in block['add_block']:
+            errors.append(self._error_message('The block {} contains a list collector a routing rule on the "add_block"'
```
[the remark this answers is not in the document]
bitdivision: Same for the following strings too.

tests/schemas/invalid/test_invalid_list_collector_with_routing.json
```diff
@@ -0,0 +1,156 @@
...
+                "type": "ListCollectorConfirmation",
+                "title": "Does anyone else live here?",
+                "answers": [
+                  {
+                    "id": "anyone-else",
+                    "mandatory": true,
+                    "type": "ListCollectorConfirmationRadio",
+                    "options": [
+                      {
+                        "label": "Yes"
```
ajmaddaford: I think that relying on an empty value do determine the add action is unclear. It's also not ideal to have new types for the block, question and answer. I've discussed this with @jonnyshaw89 and we now prefer a new block level property of `add_answer_value` -  string value that matches whatever radio answer should be directed to the add block. This would also allow us to remove the special question and answer types.

jonnyshaw89: Addressed

app/validation/validator.py
```diff
@@ -333,6 +329,11 @@ def validate_list_collector_type(self, block):
                     errors.append(self._error_message('The block {} contains a list collector without a Radio answer type'
                                                       .format(block['id'])))
 
+                if not self.options_contain_value(answer['options'], add_answer_value):
+                    errors.append(
+                        self._error_message('The block {} contains a list collector with a add_answer_value, that is not in an option'
```
ajmaddaford: The block is a list collector, rather than containing one, so this should be something like "The list collector block {} has an add_answer_value that is not present in the answer values"

app/validation/validator.py
```diff
@@ -333,6 +329,11 @@ def validate_list_collector_type(self, block):
...
+                                            .format(block['id'])))
+
         if 'routing_rules' in block['add_block']:
             errors.append(self._error_message('The block {} contains a list collector and a routing rule on the "add_block"'
```
ajmaddaford: Same as above, this should be rephrased as well.

PR #117, opened 2019-03-13
Validate Placeholder Source Ids

iwootten:
**Motivation**

It's currently possible to insert invalid references to answer ids within placeholders, for example those that reference the same block or don't exist in the schema at all.

https://trello.com/c/s9gGN5wZ/2821-placeholder-answer-source-identifier-must-be-valid

**Changes**

Validate against invalid answer ids and update current schemas that use invalid answer ids.

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -13,7 +13,7 @@ MIN_NUMBER = -999999999
 MAX_DECIMAL_PLACES = 6
 
 
-class Validator:    # pylint: disable=too-many-public-methods
+class Validator:    # pylint: disable=too-many-public-methods, too-many-lines
     def __init__(self):
         with open('schemas/questionnaire_v1.json', encoding='utf8') as schema_data:
             self.schema = load(schema_data)
@@ -24,9 +24,9 @@ class Validator:    # pylint: disable=too-many-public-methods
         :param json_to_validate: json schema to be validated
         :return: list of dictionaries containing error messages, otherwise it returns an empty list
         """
-        all_errors = {}
-
-        all_errors['schema_errors'] = self._validate_json_against_schema(json_to_validate)
+        all_errors = {
+            'schema_errors': self._validate_json_against_schema(json_to_validate)
+        }
 
         validation_errors = []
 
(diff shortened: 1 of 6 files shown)


iwootten:
Just noticed the comment on metadata needing validation too on the original card, so marking this as wip until that's completed.

app/validation/validator.py
```diff
@@ -15,7 +15,7 @@
 MAX_DECIMAL_PLACES = 6
 
 
-class Validator:    # pylint: disable=too-many-public-methods
+class Validator:    # pylint: disable=R0904, C0302
```
bitdivision: This still irks me, AFAICS pylint should allow you to use multiple named rules.

iwootten: Yes me too, as I said - I've tried but couldn't get it to work when the rules were named.

iwootten: Huh, turned out the rule name I was using was incorrect as it was named differently in the error. I've updated this now.

bitdivision: Nice.

tests/test_schema_validation.py
```diff
@@ -326,6 +326,27 @@ def test_invalid_string_transforms():
...
+        'Schema Integrity Error. Invalid answer id reference `invalid-answer0` for placeholder `simple_answer`',
+        'Schema Integrity Error. Invalid answer id reference `invalid-answer1` for placeholder `answer1`',
+        'Schema Integrity Error. Invalid metadata reference `invalid-metadata-ref` for placeholder `simple_metadata`'
+    ]
+
+    check_validation_errors(filename, expected_error_messages)
+
+
+def test_placeholder_answer_ids():
+    filename = 'schemas/valid/test_placeholder_source_ids.json'
```
jonnyshaw89: we dont need to test valid schemas do we? it happens automatically?

bitdivision: I'd forgotten about that. Yeah, it does.

bitdivision: @iwootten It creates tests for all of the valid schemas here: https://github.com/ONSdigital/eq-schema-validator/blob/7d368204b584fa12cdd3c45d40d49cfa07ac1132/tests/conftest.py

I'm not sure how to make that more discoverable?

bitdivision:
I'm happy if you want to rebase @iwootten

PR #118, opened 2019-03-14
Reduce the restriction on the format of an id

tgandrews:
Author is changing to use a short id instead of a guid for the ids as our urls are being incredibly long and cumbersome with guids. However, we are still using our ids as the entity ids in runner and this causes runner to reject them.

A short id is between 7 and 14 characters and can use any of A-Z, a-z, 0-9, - or _. The previous regex was too restrictive for these.

AFAICT there is no reason for the id regex to be so restrictive. So I have changed it which resolves the issue in author.

Please confirm this is acceptable.

### DIFF
--- a/schemas/common_definitions.json
+++ b/schemas/common_definitions.json
@@ -3,7 +3,7 @@
   "id": {
     "type": "string",
     "description": "Used to identify groups, blocks, questions and answers.",
-    "pattern": "^[a-z0-9][a-z0-9\\-]*[a-z0-9]$"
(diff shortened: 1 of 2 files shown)


tgandrews:
As discussed with @ajmaddaford this restriction is too open as they are used as urls and whilst people are still manually creating ids then we want them to be kebab case. 

The plan is to either:
1. restrict the format of ids in author 
2. change what we are sending to runner by deriving ids from content https://trello.com/c/IwhtDg7i/991-add-custom-id-fields

PR #121, opened 2019-03-21
Enfore consistent ids across variants

bitdivision:
To make the implementation of variants simpler in runner, this enforces that all question_ids are the same between each question_variant in a block. See https://github.com/ONSdigital/eq-survey-runner/pull/2021

I can't think of any situation where having multiple question_ids within a block is useful to us.

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -144,6 +144,44 @@ class Validator:    # pylint: disable=too-many-public-methods, too-many-lines
 
         return errors
 
+    def _ensure_all_variant_ids_consistent(self, block, variants):
+        """ Ensure consistency between ids in variants
+
+        - Ensure that question_ids are the same across all variants.
+        - Ensure answer_ids are the same across all variants.
+        """
+        if not variants:
+            return []
+
+        errors = []
+
+        question_ids = set()
+        answer_ids = set()
+
+        number_of_answers = set()
+
+        for variant in variants:
+            question_variant = variant['question']
+            question_ids.add(question_variant['id'])
+
+            number_of_answers.add(len(variant['question']['answers']))
+
+            if len(number_of_answers) > 1:
+                errors.append(self._error_message('Variants in block: {} contain different numbers of answers'.format(block['id'])))
+
+            for answer in question_variant['answers']:
+                answer_ids.add(answer['id'])
+
+        if len(question_ids) != 1:
+            errors.append(self._error_message(
+                'Variants contain more than one question_id for block: {}. Found ids: {}' .format(block['id'], question_ids)))
+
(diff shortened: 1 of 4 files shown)


app/validation/validator.py
```diff
@@ -144,6 +144,20 @@ def _validate_questions(self, block_or_variant, numeric_answer_ranges):
 
         return errors
 
+    def _ensure_all_variant_ids_equal(self, block, variants):
```
jonnyshaw89: Do we also need to enforce this at an `answer_id` level?

bitdivision: I hesitated on this because I think this would also require enforcing the same number of answers across variants? Maybe that use case is something we should use routing for anyway though.

jonnyshaw89: I can see a case for different answer options but not different answers. At the point we get to different answers then i think that would not be a variant.

MebinAbraham: Not sure, I can see scenarios where we would have different no. of answers?. If you had situations where a user is either shown a `from date` vs `from date` and `to date`?. 
Or a user being shown a list of checkboxes (1 answer) vs shown a mutually exclusive checkbox (2 answers), these should technically be variants and not really new question blocks?

Enforcing consistent answer ids for also won't work for mutually exclusive either, as the exclusive answer is a diff id..?

jonnyshaw89: hmm, in not sure that really fits into the idea of what variants was for. maybe we need to be more explicit on the use cases for variants vs routing?

MebinAbraham: I think its fine as long as we clearly define what is considered a variant vs a new question.

bitdivision: Agreed, just spoken to Mebin about this, we need to define the use cases for routing and variants more clearly. The original requirements for variants was to support different versions of:

- question titles
- answer guidance
- interstitials
- detail answer
- response options

Having from_date and to_date seems like a different question to me.

app/validation/validator.py
```diff
@@ -156,12 +191,20 @@ def _validate_variants(self, block, answer_ids_with_group_id, numeric_answer_ran
...
             errors.append(self._error_message('Variants contains fewer than two variants - block: {}'.format(block['id'])))
 
+        number_of_answers = set()
+
         for variant in question_variants:
             errors.extend(self._validate_questions(variant, numeric_answer_ranges))
+            number_of_answers.add(len(variant['question']['answers']))
+
+        if len(number_of_answers) > 1:
+            errors.append(self._error_message('Variants in block: {} contain different numbers of answers'.format(block['id'])))
```
ajmaddaford: It wold make more sense for this validation to be in `_ensure_all_variant_ids_consistent`

PR #122, opened 2019-03-26
Remove repeats

jonnyshaw89:
Removed repeats and relationship from validator as we no longer have this functionality in runner


schemas/questions/types/relationship.json
```diff
@@ -1,57 +0,0 @@
-{
```
iwootten: How come this has been removed? Is it a type of repeat?

jonnyshaw89: Its also been removed from runner

tests/test_schema_validation.py
```diff
@@ -43,10 +43,10 @@ def check_validation_errors(filename, expected_validation_error_messages):
 
     assert schema_errors == {}
 
-    assert len(validation_errors) == len(expected_validation_error_messages)
```
bitdivision: This will not longer ensure that there are no extra validation errors?

jonnyshaw89: yeh I switched for loop to loop over the expected error messages and removed this so it gave a better error message rather than expected 5 got 6.
Ill add this back in under the loop.

PR #126, opened 2019-04-08
Changes for list collector

bitdivision:
There are two changes here:

1) Add a remove_answer_value on the top level of the ListCollector block definition. 
    - Runner needs some way to determine the value to compare against when the remove block is posted. I did explore the option of having a custom answer type which hardcodes the answer value, e.g. `ConfirmationRadio` but this resulted in a large amount of duplication in the schema, which I can't really see any way around.
    - Perhaps another option here is just to enforce the value in the validator python, and not in the schema.
2) Enforce consistent answer_ids between list collectors
    - If multiple collectors populate a single list, they need to use the same answer_ids. 
    - Additionally, add and edit blocks must always contain the same answer_ids.

Would appreciate general feedback on point 1, as I'm not sure of the best approach.


app/validation/validator.py
```diff
@@ -329,27 +331,32 @@ def validate_skip_condition(self, skip_condition, answer_ids_with_group_id, bloc
...
-                    errors.append(self._error_message('The list collector block {} does not contain a Radio answer type'
+                    errors.append(self._error_message('The list collector add block {} does not contain a Radio answer type'
                                                       .format(block['id'])))
 
                 if not self.options_contain_value(answer['options'], add_answer_value):
                     errors.append(
                         self._error_message('The list collector block {} has an add_answer_value that is not present in the answer values'
                                             .format(block['id'])))
 
+        for add_question in add_questions:
```
MebinAbraham: How come this is duplicated?

app/validation/validator.py
```diff
@@ -329,27 +331,32 @@ def validate_skip_condition(self, skip_condition, answer_ids_with_group_id, bloc
...
+            for answer in add_question['answers']:
+                if answer['type'] != 'Radio':
+                    errors.append(self._error_message('The list collector remove block {} does not contain a Radio answer type'
+                                                      .format(block['id'])))
+
+                if not self.options_contain_value(answer['options'], remove_answer_value):
```
MebinAbraham: Worth verifying that the add and remove answer value are unique?

schemas/answers/definitions.json
```diff
@@ -1,93 +0,0 @@
-{
```
MebinAbraham: 👀 
`jsonschema.exceptions.RefResolutionError: <urlopen error [Errno 2] No such file or directory: '/Users/abrahm/Documents/projects/eq-schema-validator/schemas/answers/definitions.json'>
`

app/validation/validator.py
bitdivision: Good point, I had obviously copied but not modified this.

app/validation/validator.py
```diff
@@ -1082,3 +1115,18 @@ def _get_answers_with_parent_ids(json_to_validate):
...
         return answers
+
+    @staticmethod
+    def _get_all_questions_for_block(block):
+        """ Get all questions on a block including variants"""
+        questions = []
+
+        for variant in block.get('question_variants', []):
+            if 'question' in variant:
+                questions.append(variant.get('question', {}))
```
MebinAbraham: Me being picky but at this point the default would never be hit anyway?.
So `questions.append(variant['question'])`?

app/validation/validator.py
```diff
@@ -347,39 +352,77 @@ def validate_skip_condition(self, skip_condition, answer_ids_with_group_id, bloc
...
+
+    def validate_list_collector_answer_ids(self, block):
+        """
+        - Ensure that answer_ids on add blocks match between all blocks that populate a single list.
+        - Enforce the same answer_ids on add and edit sub-blocks
+        """
+        errors = []
+        list_name = block['populates_list']
+
+        add_block_questions = self._get_all_questions_for_block(block['add_block'])
```
MebinAbraham: All references by key that is not using `.get()` isn't validating the schema since it fails on key error. It is mandatory in the schemas but it doesn't tell you that until it does our python check also.

```python
File "/Users/abrahm/Documents/projects/eq-schema-validator/app/validation/validator.py", line 358, in validate_list_collector_type
    remove_questions = self._get_all_questions_for_block(block['remove_block'])
KeyError: 'remove_block'
```

app/validation/validator.py
```diff
@@ -348,39 +353,77 @@ def _validate_skip_condition(self, skip_condition, answer_ids_with_group_id, blo
...
-        if 'question_variants' in block:
-            for question in block['question_variants']:
-                questions.append(question)
-
-        for question in questions:
-            for answer in question['answers']:
-                if answer['type'] != 'Radio':
+        collector_questions = self._get_all_questions_for_block(block)
+        remove_questions = self._get_all_questions_for_block(block['remove_block'])
+        add_answer_value = block['add_answer']['value']
```
MebinAbraham: Need to also validate that the `block['add_answer']['id']` is an `id` that is present within the block as runner will 500 with a KeyError if they mismatch.

app/validation/validator.py
```diff
@@ -348,39 +353,77 @@ def _validate_skip_condition(self, skip_condition, answer_ids_with_group_id, blo
...
+        remove_answer_value = block['remove_answer']['value']
```
MebinAbraham: Same comment about validating the id of add answer.

app/validation/validator.py
MebinAbraham: Still an issue.

app/validation/validator.py
```diff
@@ -348,39 +356,97 @@ def _validate_skip_condition(self, skip_condition, answer_ids_with_group_id, blo
...
+            errors.append(self._error_message(
+                'Remove answer reference uses id not found in remove_block: {}'.format(block["remove_answer"]["id"]))
+            )
 
-        for question in questions:
-            for answer in question['answers']:
-                if answer['type'] != 'Radio':
+        return errors
+
+    def _validate_list_collector_type(self, block):  # noqa: C901  pylint: disable=too-complex, too-many-locals
```
MebinAbraham: This is validating more than just the type, we could move the checking of `add_answer_value = block['add_answer']['value']` etc into the new `_validate_list_answer_references`.

bitdivision: I've changed the name of the top level method to `_validate_list_collector` since the type doesn't have anything to do with it.

I'm reluctant to move the rest of the validation out since it relies on the data being built up in that function whereas validating the ids is fairly self contained.

schemas/blocks/list_collector.json
```diff
@@ -1,5 +1,21 @@
...
+        "description": "The id of an answer to reference"
+      },
+      "value": {
+        "type": [
+          "string",
+          "integer"
+        ],
+        "description": "The value of the answer which should be matched"
+      }
+    }
```
MebinAbraham:
```suggestion
    },
    "additionalProperties": false
```

bitdivision: Good spot.

schemas/blocks/list_collector.json
```diff
@@ -1,5 +1,22 @@
...
+      },
+      "value": {
+        "type": [
+          "string",
+          "integer"
+        ],
+        "description": "The value of the answer which should be matched"
+      }
+    },
+    "additionalProperties": false
```
MebinAbraham: Also should use 
```json
"required": [ "id", "value"],
```

app/validation/validator.py
```diff
@@ -98,7 +101,13 @@ def _validate_blocks(self, json_to_validate, section, group, all_groups, answers
                 errors.extend(self._validate_calculated_summary_type(block, answers_with_parent_ids))
 
             if block['type'] == 'ListCollector':
-                errors.extend(self._validate_list_collector_type(block))
+                try:
+                    errors.extend(self._validate_list_collector(block))
+                except KeyError as e:
+                    errors.append(f'Missing key in list collector: {e}')
+
+            if block['type'] in ['ListAddQuestion', 'ListEditQuestion', 'ListRemoveQuestion']:
```
MebinAbraham:
```suggestion
            elif block['type'] in ['ListAddQuestion', 'ListEditQuestion', 'ListRemoveQuestion']:
```

MebinAbraham:
Happy with everything else.

PR #127, opened 2019-04-08
Ability to define multiple values in when condition

MebinAbraham:
Ability to define a list of values in when condition.
Introduced these new conditions:
1. `equals any`
2. `not equals any`
3. `contains any`
4. `contains all`

Needed for https://github.com/ONSdigital/eq-survey-runner/pull/2041

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -1,9 +1,10 @@
 import os
-import re
 import pathlib
-from json import load
+import re
 from collections import defaultdict
 from datetime import datetime
+from json import load
+
 from dateutil.relativedelta import relativedelta
 from jsonschema import SchemaError, RefResolver, validate, ValidationError
 from jsonschema.exceptions import best_match
@@ -13,7 +14,7 @@ MIN_NUMBER = -999999999
 MAX_DECIMAL_PLACES = 6
 
 
-class Validator:    # pylint: disable=too-many-public-methods, too-many-lines
+class Validator:  # pylint: disable=too-many-lines
     def __init__(self):
         with open('schemas/questionnaire_v1.json', encoding='utf8') as schema_data:
             self.schema = load(schema_data)
@@ -31,7 +32,7 @@ class Validator:    # pylint: disable=too-many-public-methods, too-many-lines
         validation_errors = []
 
         validation_errors.extend(self._validate_schema_contain_metadata(json_to_validate))
-        validation_errors.extend(self.validate_duplicates(json_to_validate))
+        validation_errors.extend(self._validate_duplicates(json_to_validate))
 
         numeric_answer_ranges = {}
         answers_with_parent_ids = self._get_answers_with_parent_ids(json_to_validate)
@@ -46,7 +47,7 @@ class Validator:    # pylint: disable=too-many-public-methods, too-many-lines
                 validation_errors.extend(self._validate_routing_rules(group, all_groups, answers_with_parent_ids))
 
                 for skip_condition in group.get('skip_conditions', []):
-                    validation_errors.extend(self.validate_skip_condition(skip_condition, answers_with_parent_ids, group))
+                    validation_errors.extend(self._validate_skip_condition(skip_condition, answers_with_parent_ids, group))
 
                 validation_errors.extend(
                     self._validate_blocks(json_to_validate,
@@ -64,11 +65,11 @@ class Validator:    # pylint: disable=too-many-public-methods, too-many-lines
     def _validate_routing_rules(self, group, all_groups, answers_with_parent_ids):
         errors = []
 
-        errors.extend(self.validate_routing_rules_default(group.get('routing_rules', []), group))
+        errors.extend(self._validate_routing_rules_default(group.get('routing_rules', []), group))
         for rule in group.get('routing_rules', []):
-            errors.extend(self.validate_schema_routing_rule_routes_to_valid_target(group['blocks'], 'block', rule))
-            errors.extend(self.validate_schema_routing_rule_routes_to_valid_target(all_groups, 'group', rule))
-            errors.extend(self.validate_routing_rule(rule, answers_with_parent_ids, group))
+            errors.extend(self._validate_schema_routing_rule_routes_to_valid_target(group['blocks'], 'block', rule))
+            errors.extend(self._validate_schema_routing_rule_routes_to_valid_target(all_groups, 'group', rule))
+            errors.extend(self._validate_routing_rule(rule, answers_with_parent_ids, group))
 
         return errors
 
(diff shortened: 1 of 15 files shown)


MebinAbraham:
@ajmaddaford Have split up the when rules into separate schemas. However, I have grouped them together to what I think makes sense. Having a schema for every when rules didn't seem necessary. Have a look and let me know.

schemas/when_rule/definitions.json
```diff
@@ -0,0 +1,87 @@
...
+      },
+      "value": {
+        "allOf": [
+          {
+            "$ref": "#/comparison_value"
+          },
+          {
+            "pattern": "^(now|\\d{4}\\-(0?[1-9]|1[012])(|\\-(0?[1-9]|[12][0-9]|3[01])))$"
+          }
+        ]
```
ajmaddaford: `allOf` with `comparison_value` and `pattern` doesn't make sense as the pattern enforces a string and comparison value can be a string, boolean or integer.

app/validation/validator.py
```diff
@@ -491,38 +487,38 @@ def _validate_answer_ids_present_in_schema(self, when, answer_ids_with_group_id,
...
+        checkbox_exclusive_conditions = ('contains any', 'contains all')
+        invalid_checkbox_conditions = ('equals any', 'not equals any')
+        conditions_requiring_list_match_values = (*invalid_checkbox_conditions, *checkbox_exclusive_conditions)
+
+        if condition in conditions_requiring_list_match_values and not when.get('values'):
+            errors.append(self._error_message(f'Property `values` is missing/empty for when condition `{condition}` with answer id `{when["id"]}`'))
+
+        answer_type = answer_ids_with_group_id[when['id']]['answer']['type'] if 'id' in when else None
+        if answer_type == 'Checkbox':
+            if condition in invalid_checkbox_conditions:
```
ajmaddaford: I think all conditions other than `checkbox_exclusive_conditions` are invalid for a checkbox source answer, so this could be `if condition not in checkbox_exclusive_conditions`.

app/validation/validator.py
```diff
@@ -491,38 +487,38 @@ def _validate_answer_ids_present_in_schema(self, when, answer_ids_with_group_id,
...
-        else:  # We know we have a 'answer_ids' key so can do validation against it
-            if when['condition'] in ('contains', 'not contains'):
-                errors.append(self._error_message('The condition "{}" is not valid for an answer_count based "when" clause'
-                                                  .format(when['condition'])))
-            if len(when['answer_ids']) != len(set(when['answer_ids'])):
-                errors.append(self._error_message('Duplicate answer ids found within {} clause'.format(referenced_id)))
+        condition = when['condition']
+        checkbox_exclusive_conditions = ('contains any', 'contains all')
+        invalid_checkbox_conditions = ('equals any', 'not equals any')
+        conditions_requiring_list_match_values = (*invalid_checkbox_conditions, *checkbox_exclusive_conditions)
```
ajmaddaford: The validation for checkbox conditions and conditions requiring list match values should be more clearly separated. I think making `conditions_requiring_list_match_values` an explicit list (rather than trying to make use of the checkbox conditions) and moving it to after the checkbox validation is done will help.

schemas/when_rule/definitions.json
```diff
@@ -0,0 +1,87 @@
...
+      "meta": {
+        "$ref": "#/metadata_identifier"
+      },
+      "value": {
+        "allOf": [
+          {
+            "$ref": "#/comparison_value"
+          },
+          {
+            "pattern": "^(now|\\d{4}\\-(0?[1-9]|1[012])(|\\-(0?[1-9]|[12][0-9]|3[01])))$"
```
ajmaddaford: There is already some regex used for dates in `date_value` in the string interpolation definitions schema. We should consolidate these if appropriate.

schemas/when_rule/definitions.json
```diff
@@ -0,0 +1,87 @@
...
+    "description": "The value to compare against"
+  },
+  "comparison_values": {
+    "type": "array",
+    "description": "An array of values to compare against",
+    "minItems": 1
+  },
+  "comparison_identifier": {
+    "type": "string",
+    "description": "The id of an answer from which to obtain the value and compared to"
```
ajmaddaford: compared -> compare

schemas/when_rule/definitions.json
MebinAbraham: I've now reused `date_value` from string_interpolation.

schemas/when_rule/definitions.json
MebinAbraham: I've reused the `date_value` and seem to work fine, although this pattern doesn't seem to match the same.

app/validation/validator.py
MebinAbraham: Fair, done 👍

app/validation/validator.py
```diff
@@ -491,38 +487,37 @@ def _validate_answer_ids_present_in_schema(self, when, answer_ids_with_group_id,
...
+        if answer_type == 'Checkbox':
+            if condition not in checkbox_exclusive_conditions:
+                errors.append(self._error_message(f'The condition`{condition}` cannot be used with `Checkbox` answer type.'))
+        elif condition in checkbox_exclusive_conditions:
+            errors.append(
+                self._error_message(f'The condition`{condition}` can only be used with `Checkbox` answer types. Found answer type: {answer_type}'))
+
+        conditions_requiring_list_match_values = ('equals any', 'not equals any', 'contains any', 'contains all')
+        if condition in conditions_requiring_list_match_values and not when.get('values'):
+            errors.append(self._error_message(f'Property `values` is missing/empty for when condition `{condition}` with answer id `{when["id"]}`'))
```
ajmaddaford: I don't think this validation is as simple as this. The schema allows for `values` or `comparison_id` in compare_to_list.json. If that's correct we should change this to validate that comparison_id is the id of an answer that is a list i.e. a checkbox answer.

MebinAbraham: As discussed have added certain checks without being too exhaustive. Have a look and let me know. Thanks

app/validation/validator.py
MebinAbraham: I have now defined the `valid` *(acceptable)* checkbox conditions and check that only those are used when the answer type is `Checkbox`. And that all other answer types do not use checkbox exclusive conditions.

app/validation/validator.py
```diff
@@ -497,48 +496,65 @@ def _validate_answer_ids_present_in_schema(self, when, answer_ids_with_group_id,
...
+        valid_checkbox_conditions = ('contains any', 'contains all', 'contains', 'not contains', 'set', 'not set')
         answer_type = answer_ids_with_group_id[when['id']]['answer']['type'] if 'id' in when else None
 
         if answer_type == 'Checkbox':
-            if condition not in checkbox_exclusive_conditions:
-                errors.append(self._error_message(f'The condition`{condition}` cannot be used with `Checkbox` answer type.'))
-        elif condition in checkbox_exclusive_conditions:
+            if condition not in valid_checkbox_conditions:
+                errors.append(self._error_message(f'The condition `{condition}` cannot be used with `Checkbox` answer type.'))
+        elif condition in ('contains any', 'contains all', 'contains', 'not contains'):
```
ajmaddaford: I know this is how it was originally, but now that there are more conditions in the list I think it would make sense to have:

```
checkbox_exclusive_conditions = ['contains any', 'contains all', 'contains', 'not contains']
all_checkbox_conditions = checkbox_exclusive_conditions + ['set', 'not set']
```

app/validation/validator.py
```diff
@@ -497,48 +496,65 @@ def _validate_answer_ids_present_in_schema(self, when, answer_ids_with_group_id,
...
+        """
+        errors = []
+        condition, answer_id = when['condition'], when['id']
         conditions_requiring_list_match_values = ('equals any', 'not equals any', 'contains any', 'contains all')
-        if condition in conditions_requiring_list_match_values and not when.get('values'):
-            errors.append(self._error_message(f'Property `values` is missing/empty for when condition `{condition}` with answer id `{when["id"]}`'))
+
+        if condition in conditions_requiring_list_match_values and not when.get('values', when.get('comparison_id')):
+            errors.append(self._error_message(
+                f'Either the properties `values` or `comparison_id` is missing/empty for when condition `{condition}` with answer id `{answer_id}`'))
```
ajmaddaford: Isn't this tested by the schema?

app/validation/validator.py
```diff
@@ -497,48 +496,65 @@ def _validate_answer_ids_present_in_schema(self, when, answer_ids_with_group_id,
...
+
+        if 'comparison_id' in when:
+            errors.extend(
+                self._validate_comparison_id_in_when_rule(when, answer_ids_with_group_id, referenced_id, conditions_requiring_list_match_values))
```
ajmaddaford: It's not clear that this method is also validating comparison_id for non list match values as it's named `_validate_conditions_requiring_list_match_values_are_valid_in_when_rule`.

It would be better to call `_validate_comparison_id_in_when_rule` first and from there call this method, or just move that logic into that method.

app/validation/validator.py
```diff
@@ -491,63 +486,79 @@ def _validate_answer_ids_present_in_schema(self, when, answer_ids_with_group_id,
...
+        condition, answer_id = when['condition'], when['id']
+        conditions_requiring_list_match_values = ('equals any', 'not equals any', 'contains any', 'contains all')
 
-        if comparison_type != id_type:
-            errors.append(self._error_message('The answers used as comparison_id "{}" and answer_id "{}" in the '
-                                              '"when" clause for {} have different types'
-                                              .format(when['comparison_id'], when['id'], referenced_id)))
+        if condition in conditions_requiring_list_match_values and not when.get('values', when.get('comparison_id')):
+            errors.append(self._error_message(
+                f'Either the properties `values` or `comparison_id` is missing/empty for when condition `{condition}` with answer id `{answer_id}`'))
```
iwootten: Isn't it possible to determine which of these are missing and provide a more specific error?

MebinAbraham: This is being removed as this is caught by schema definition anyway.

app/validation/validator.py
MebinAbraham: Changed back.

app/validation/validator.py
MebinAbraham: Was only kept for better error messaging but remove now.

app/validation/validator.py
MebinAbraham: Have removed `_validate_conditions_requiring_list_match_values_are_valid_in_when_rule` as JSON schema takes care of it anyway and kept `_validate_comparison_id_in_when_rule`.

PR #128, opened 2019-04-09
Validate variants question type and answer type

JamesGardiner:
Add a schema validation step that checks question types and answer types within blocks are identical, including across variants.

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -142,11 +142,13 @@ class Validator:    # pylint: disable=too-many-public-methods, too-many-lines
 
         return errors
 
-    def _ensure_all_variant_ids_consistent(self, block, variants):
-        """ Ensure consistency between ids in variants
+    def _ensure_relevant_variant_fields_are_consistent(self, block, variants):
+        """ Ensure consistency between relevant fields in variants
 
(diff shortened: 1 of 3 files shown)


app/validation/validator.py
```diff
@@ -169,14 +174,23 @@ def _ensure_all_variant_ids_consistent(self, block, variants):
...
 
         if len(question_ids) != 1:
             errors.append(self._error_message(
                 'Variants contain more than one question_id for block: {}. Found ids: {}' .format(block['id'], question_ids)))
 
+        if len(question_types) != 1:
+            errors.append(self._error_message(
+                'Variants have more than one question type for block: {}. Found types: {}'.format(block['id'], question_types)))
+
+        if len(answer_types) != 1:
```
bitdivision: I'm not sure this will work if a single variant has multiple answers with different types?

e.g. all variants have two answers with types number and textfield.

MebinAbraham: Good point, certainly possible, but I don't think I have ever seen a question with 2 different answer types.

bitdivision:
@MebinAbraham Make sure you're happy.

PR #130, opened 2019-05-14
Update metadata schema to use `type` not `validator`

bitdivision:
In combination with: https://github.com/ONSdigital/eq-survey-runner/pull/2076

Adds the schema definitions for metadata. Uses `type` rather than `validator` and allows length validation and optional fields.

### DIFF
--- a/schemas/questionnaire_v1.json
+++ b/schemas/questionnaire_v1.json
@@ -62,21 +62,40 @@
             "type": "string",
             "pattern": "^[a-zA-Z0-9-_]+$"
           },
-          "validator": {
+          "type": {
             "type": "string",
             "enum": [
               "boolean",
               "date",
               "string",
-              "optional_string",
-              "uuid"
+              "uuid",
+              "url"
             ]
+          },
+          "min_length": {
+            "type": "integer",
+            "description": "The minimum length of a string field",
+            "minimum": 1
+          },
+          "max_length": {
+            "type": "integer",
+            "description": "The maximum length of a string field",
+            "minimum": 1
+          },
+          "length": {
+            "type": "integer",
+            "description": "The length of a string field must be equal to this",
+            "minimum": 0
+          },
+          "optional": {
+            "description": "Whether an error should be raised if this field is not provided.",
+            "type": "boolean"
           }
         },
         "additionalProperties": false,
         "required": [
           "name",
-          "validator"
+          "type"
         ]
       }
     },
--- a/tests/schemas/invalid/test_invalid_answer_comparison_id.json
+++ b/tests/schemas/invalid/test_invalid_answer_comparison_id.json
@@ -10,15 +10,15 @@
   "metadata": [
     {
       "name": "user_id",
-      "validator": "string"
+      "type": "string"
     },
     {
       "name": "period_id",
-      "validator": "string"
+      "type": "string"
     },
     {
       "name": "ru_name",
-      "validator": "string"
+      "type": "string"
     }
   ],
   "sections": [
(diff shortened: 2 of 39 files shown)


schemas/questionnaire_v1.json
```diff
@@ -71,12 +71,33 @@
               "optional_string",
```
ajmaddaford: - `optional_string` needs to go
- `url` needs to be added

schemas/questionnaire_v1.json
```diff
@@ -71,12 +71,33 @@
...
+                "type": "number",
+                "description": "The minimum length of a string field"
+              },
+              "max": {
+                "type": "number",
+                "description": "The maximum length of a string field"
+              },
+              "equal": {
+                "type": "number",
+                "description": "The length of a string field must be equal to this"
```
ajmaddaford: I don't see the need to nest properties under `length`. If we were to extend the validation to numbers it's not clear what the nesting property would be (`size` possibly, but that looks a bit weird). Also, using `min_length` and `max_length` would more closely relate to json schema rather than marshmallow?

schemas/questionnaire_v1.json
```diff
@@ -62,21 +62,37 @@
...
               "date",
               "string",
-              "optional_string",
-              "uuid"
+              "uuid",
+              "url"
             ]
+          },
+          "min_length": {
+            "type": "number",
```
MebinAbraham: Use integer instead?
```suggestion
            "type": "integer",
```

schemas/questionnaire_v1.json
```diff
@@ -62,21 +62,37 @@
...
+            "description": "The minimum length of a string field"
+          },
+          "max_length": {
+            "type": "number",
```
MebinAbraham: Use integer instead?
```suggestion
            "type": "integer",
```

schemas/questionnaire_v1.json
```diff
@@ -62,21 +62,37 @@
...
+            "description": "The maximum length of a string field"
+          },
+          "length": {
+            "type": "number",
```
MebinAbraham: Use integer instead?
```suggestion
            "type": "integer",
```

schemas/questionnaire_v1.json
MebinAbraham: Worth defining: `"minimum"`?

MebinAbraham: Use integer instead?
```suggestion
            "type": "integer",
```

schemas/questionnaire_v1.json
MebinAbraham: Use integer instead?
```suggestion
            "type": "integer",
```

schemas/questionnaire_v1.json
```diff
@@ -62,21 +62,37 @@
...
+            "description": "The length of a string field must be equal to this"
+          },
+          "optional": {
```
MebinAbraham: Since JSON and Marshmello both use required, would it be better if we also used `required` instead of `optional`?

bitdivision: I prefer optional, since the field itself is optional. But happy to change if that's the consensus.

Coupling to marshmallow shouldn't really be an argument for it though

schemas/questionnaire_v1.json
```diff
@@ -62,21 +62,38 @@
...
+              "url"
             ]
+          },
+          "min_length": {
+            "type": "integer",
+            "description": "The minimum length of a string field",
+            "minimum": 1
+          },
+          "max_length": {
+            "type": "integer",
```
MebinAbraham: "minimum" here also and maybe if we want to define a maximum?

schemas/questionnaire_v1.json
```diff
@@ -62,21 +62,38 @@
...
+            "description": "The maximum length of a string field"
+          },
+          "length": {
+            "type": "integer",
```
MebinAbraham: minimum here?

PR #131, opened 2019-05-20
Remove block title and description

jonnyshaw89:
🤷‍♂ What it says on the tin


tests/schemas/invalid/test_invalid_date_range_period.json
```diff
@@ -2,74 +2,59 @@
...
-                    "days": 23,
-                    "years": 1
-                  }
-                },
-                "title": "Enter Date Range",
-                "type": "DateRange"
-              },
-              "title": "Date Range",
-              "type": "Question"
+  "sections": [{
```
bitdivision: How has this changed? Is yarn format not able to correct it?

schemas/blocks/confirmation_question.json
```diff
@@ -6,15 +6,6 @@
       "id": {
         "$ref": "../common_definitions.json#/id"
       },
-      "section_number": {
```
bitdivision: 👍

tests/schemas/invalid/test_invalid_date_range_period.json
bitdivision: I see. It was broken

jonnyshaw89: yeh this is the output from the format command

ajmaddaford:
After discussions with @jonnyshaw89 we've decided to not use `body`. It didn't look right and the way we were using it was closer to the html definition of a body rather than the more general [body of text](https://en.wikipedia.org/wiki/Body_text) definition. 

We're going to rename the `content` elements that are a list of arbitrary content (headings, paragraphs and lists) to `contents` to differentiate it from the content element at the block level. Something like:

```
{
  "id": "content-block",
  "type": "Interstitial",
  "content": {
    "title": "My content block",
    "contents": [
      {
        "description": "This is a nice block"
      },
      {
        "list": [
          "item 1",
          "item 2"
        ]
      }
    ]
  }
}
```

schemas/blocks/introduction.json
```diff
@@ -6,12 +6,6 @@
       "id": {
         "$ref": "../common_definitions.json#/id"
       },
-      "title": {
-        "$ref": "../string_interpolation/definitions.json#/string_with_placeholders"
-      },
```
ajmaddaford: Title is probably still be needed here as otherwise there is no title for the page.

schemas/common_definitions.json
```diff
@@ -18,20 +18,21 @@
...
-          "$ref": "common_definitions.json#/content"
+        "contents": {
+          "$ref": "common_definitions.json#/contents"
         }
       },
       "additionalProperties": false
     }
   },
   "guidance": {
+    "description": "Answer guidance",
```
ajmaddaford: We should rename this definition from `guidance` to `answer_guidance` to differentiate it from question guidance.

schemas/common_definitions.json
```diff
@@ -42,12 +43,25 @@
...
       },
-      "content": {
-        "$ref": "common_definitions.json#/content"
+      "contents": {
+        "$ref": "common_definitions.json#/contents"
       }
     }
   },
   "content": {
+    "description": "Allows customisation of guidance text.",
```
ajmaddaford: This description is wrong. 

I'm not sure this definition should be in common definitions. It's only used in `confirmation.json` and `interstitial.json` and having it here adds to the confusion around `contents`. It may be better to accept a small amount of duplication and put it in both `confirmation.json` and `interstitial.json`.

schemas/common_definitions.json
```diff
@@ -42,12 +43,25 @@
...
+    "type": "object",
+    "properties": {
+      "title": {
+        "type": "string",
+        "description": "The H1 title on the page."
```
ajmaddaford: The H1 is implementation specific, it would be better to say something like "main" or "top-level" heading on the page.

schemas/common_definitions.json
```diff
@@ -42,12 +43,25 @@
...
+      },
+      "contents": {
+        "$ref": "common_definitions.json#/contents"
+      }
+    }
+  },
+  "contents": {
```
ajmaddaford: The description here is wrong, `contents` is used anywhere we need flexible content (headings, paragraphs, lists).

schemas/questions/types/general.json
```diff
@@ -25,7 +22,7 @@
         "type": "object",
         "properties": {
           "content": {
-            "$ref": "../../common_definitions.json#/content"
+            "$ref": "../../common_definitions.json#/contents"
```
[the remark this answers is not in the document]
bitdivision: Old comment?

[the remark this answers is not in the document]
ajmaddaford: No, the property should be `contents` rather than `content`.

schemas/answers/definitions.json
```diff
@@ -91,5 +91,22 @@
...
+                "type": "string",
+                "description": "The text that is used for the 'Show guidance' link."
+            },
+            "hide_guidance": {
+                "type": "string",
+                "description": "The text that is used for the 'Hide guidance' link."
+            },
+            "contents": {
+                "$ref": "../common_definitions.json#/contents"
+            }
```
MebinAbraham: It's worth using `"additionalProperties": false,`

schemas/blocks/definitions.json
```diff
@@ -0,0 +1,63 @@
...
+    "type": "object",
+    "properties": {
+      "title": {
+        "type": "string",
+        "description": "The top-level heading on the page"
+      },
+      "contents": {
+        "$ref": "../common_definitions.json#/contents"
+      }
+    }
```
MebinAbraham: Same here, worth adding `"additionalProperties": false,`
We should also enforce the `required` keys to prevent invalid schemas.

bitdivision:
The preview_content in introduction did not have `additionalProperties: false` and we were using a key called 'questions' inside it. I've defined the schema for this as per runner, but it would be worth making sure we want it as is @ajmaddaford @MebinAbraham @iwootten .

schemas/blocks/definitions.json
```diff
@@ -12,7 +12,8 @@
         "$ref": "../common_definitions.json#/contents"
       }
     },
-    "additionalProperties": false
+    "additionalProperties": false,
+    "required": ["contents"]
```
MebinAbraham: Should it not be one of `title` or `contents`?

bitdivision: I think the content title should always be required? 

And it probably should be possible to have a content block with only a title?

bitdivision: So I've gone with required: title now.

ajmaddaford:
> The preview_content in introduction did not have `additionalProperties: false` and we were using a key called 'questions' inside it. I've defined the schema for this as per runner, but it would be worth making sure we want it as is @ajmaddaford @MebinAbraham @iwootten .

For now, yes. The preview_content needs to be reviewed and could be replaced with something more in line with standard `contents`.

PR #132, opened 2019-05-24
Allow default value for Radio answer and fix date comparision error

MebinAbraham:
Allows the use of defaults for 'Radio' answer types as on the census, if for proxy they want to treat unanswered the same as if the user answered 'Yes', so instead of duplication the rule 100+ times, this solves the problem if we are happy modifying the user's answer.

Have also fixed an issue with `_get_offset_date_value` which was causing validation errors where it was possible to compare a *string* object with a *datetime* object when one of min/max had an offset defined but the other did not.

A test did not seem necessary for these changes as one is fixing an *Exception* and the other is done by JSON schema.

Needed for: https://github.com/ONSdigital/eq-survey-runner/pull/2098

### DIFF
--- a/schemas/answers/radio.json
+++ b/schemas/answers/radio.json
@@ -36,6 +36,10 @@
       "mandatory": {
         "type": "boolean"
       },
+      "default": {
+        "type": "string",
+        "description": "Default value if no answer given"
+      },
       "validation": {
         "type": "object",
         "properties": {
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -1116,7 +1116,7 @@ class Validator:  # pylint: disable=too-many-lines
 
         if 'offset_by' in answer_min_or_max:
             offset = answer_min_or_max['offset_by']
-            value = self._get_relative_date(value, offset)
+            value = self._get_relative_date(value, offset).strftime('%Y-%m-%d')
 
         return value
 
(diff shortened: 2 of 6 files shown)


bitdivision:
Is the lack of an answer significant? i.e. Can we actually use defaults on the census?

MebinAbraham:
> Is the lack of an answer significant? i.e. Can we actually use defaults on the census?

This was done to reduce the duplicated rules in the schemas, I'll check with Charlotte about defaults as I guess we only used them on number answers previously which are be used with confirmation questions for a reason?.

ajmaddaford:
@bitdivision @MebinAbraham I've spoken with @pricem14pc and @charlotte-hirst about this. Marking up the answer in the schema with a default makes sense, but the runner implementation needs to change so we don't persist the default in the answer store. This will give us the desired behaviour that the question remains unanswered to the user and isn't submitted downstream. I've raised another card in the tech backlog to change the default answer behaviour in runner.

schemas/answers/radio.json
```diff
@@ -36,6 +36,10 @@
       "mandatory": {
         "type": "boolean"
       },
+      "default": {
+        "type": "string",
+        "description": "Default value on POST if none given"
```
ajmaddaford: This shouldn't mention POST as that's implementation specific.

PR #134, opened 2019-06-24
Add lists to when rules

ajmaddaford:
Adds lists to when rules to enable routing logic based on the number of items in a list.

### DIFF
--- a/schemas/common_definitions.json
+++ b/schemas/common_definitions.json
@@ -75,10 +75,14 @@
     "type": "array",
     "minItems": 1,
     "uniqueItems": true,
-    "properties": {
-      "when": {
-        "$ref": "when_rule/definitions.json#/when"
-      }
+    "items": {
+      "type": "object",
+      "properties": {
+        "when": {
+          "$ref": "when_rule/definitions.json#/when"
+        }
+      },
+      "additionalProperties": false
     },
     "required": [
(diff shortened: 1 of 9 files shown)


schemas/common_definitions.json
```diff
@@ -75,9 +75,12 @@
...
     "uniqueItems": true,
-    "properties": {
-      "when": {
-        "$ref": "when_rule/definitions.json#/when"
+    "items": {
+      "type": "object",
+      "properties": {
+        "when": {
+          "$ref": "when_rule/definitions.json#/when"
+        }
```
MebinAbraham: `"additionalProperties": False`?

schemas/when_rule/compare_to_list_count.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "type": "object",
+  "description": "Comparison against a list count",
+  "properties": {
+    "list": {
+      "type": "string",
```
MebinAbraham: Personally, I think it is beneficial to check that the list specified here does actually exist within the schema?

ajmaddaford: I've added a check for this in the latest commit

PR #135, opened 2019-06-28
Feature: Hub & Spoke

MebinAbraham:
Allows the use of a top-level object for the hub and spoke feature.
Needed for: https://github.com/ONSdigital/eq-survey-runner/pull/2142

For now, the hub is defined as a top-level schema object but this is subject to change once https://trello.com/c/BJPC98fW/3086-move-final-summary-block-out-of-sections is implemented. We should potentially have a single selection of the submission type for a schema ie Summary/Confirmation/Hub.

### How to review 
Test against runner branch.

### DIFF
--- a/schemas/questionnaire_v1.json
+++ b/schemas/questionnaire_v1.json
@@ -137,6 +137,19 @@
     "messages": {
       "$ref": "common_definitions.json#/messages"
     },
+    "hub": {
+      "type": "object",
+      "properties": {
+        "enabled": {
+          "description": "Whether or not the hub and spoke pattern is enabled for this questionnaire.",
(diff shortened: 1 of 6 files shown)


schemas/questionnaire_v1.json
```diff
@@ -137,6 +137,19 @@
     "messages": {
       "$ref": "common_definitions.json#/messages"
     },
+    "hub": {
```
bitdivision: We should probably use this in one of the test schemas.

MebinAbraham: Added valid and invalid schemas.

PR #138, opened 2019-07-03
Adds definitions for relationship collector

JamesGardiner:
#### This PR adds new schema definitions to support the collection of relationships in surveys.

There are two new definitions:
1) `/schemas/blocks/relationships_collector.json` uses a new enum, `RelationshipsCollector`, and uses a new mandatory `list` property. This is the id of the list that contains the list items for which relationships need to be collected.
2) `/schemas/answers/relationships.json` defines new answer options properties `relationships_title` sand `relationships_playback` that are unique to and mandatory for relationships answers.

### DIFF
--- a/schemas/blocks/relationship_collector.json
+++ b/schemas/blocks/relationship_collector.json
@@ -1,0 +1,51 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "block": {
+    "type": "object",
+    "properties": {
+      "id": {
+        "$ref": "../common_definitions.json#/id"
+      },
+      "type": {
+        "type": "string",
+        "enum": ["RelationshipCollector"]
+      },
+      "title": {
+        "type": "string"
+      },
+      "for_list": {
+        "type": "string"
+      },
+      "question": {
+        "$ref": "../questions/definitions.json#/question"
+      },
+      "question_variants": {
+        "$ref": "definitions.json#/question_variants"
+      },
+      "routing_rules": {
+          "$ref": "../common_definitions.json#/routing_rules"
+      },
+      "skip_conditions": {
+          "$ref": "../common_definitions.json#/skip_conditions"
+      }
+    },
+    "additionalProperties": false,
+    "required": [
+      "id",
+      "type",
+      "for_list"
+    ],
+    "oneOf": [
+      {
+        "required": [
+          "question"
+        ]
+      },
+      {
+        "required": [
+          "question_variants"
+        ]
+      }
+    ]
+  }
+}
--- a/tests/schemas/invalid/test_invalid_relationship_list_doesnt_exist.json
+++ b/tests/schemas/invalid/test_invalid_relationship_list_doesnt_exist.json
@@ -1,0 +1,68 @@
+{
+  "mime_type": "application/json/ons/eq",
+  "schema_version": "0.0.1",
+  "data_version": "0.0.3",
+  "survey_id": "0",
+  "title": "Test Relationships",
+  "theme": "default",
+  "description": "A questionnaire to test capturing of relationships.",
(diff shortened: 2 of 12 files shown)


schemas/blocks/relationship_collector.json
```diff
@@ -0,0 +1,51 @@
...
+        "$ref": "../common_definitions.json#/id"
+      },
+      "type": {
+        "type": "string",
+        "enum": ["RelationshipCollector"]
+      },
+      "title": {
+        "type": "string"
+      },
+      "list": {
```
MebinAbraham: Maybe `list_name` or something like that?. I think we try to refrain using python built-in types so the python codes are not confusing etc.

tests/schemas/invalid/test_invalid_relationship_no_list_specified.json
```diff
@@ -0,0 +1,67 @@
+{
```
MebinAbraham: So this schema ensures that a list key is specified but we should ensure that if a list is specified, then that value of the list is actually a valid list?. Because if I pass a non-exisitent list to the `list` the check passes.

JamesGardiner:
@ajmaddaford What do you think of @MebinAbraham's comment above? I have no strong viewpoint on `for_list` vs `uses_list`.

ajmaddaford:
@JamesGardiner @MebinAbraham I think we should use `for_list` for both (list collectors and relationship collectors).

JamesGardiner:
@ajmaddaford In that case I'll leave this card as is. Will need a new card to change the ListCollector

bambrp:
> @ajmaddaford In that case I'll leave this PR as is. Will need a new card to change the ListCollector

Have created a card in backlog: https://trello.com/c/or6AdKuc/3133-listcollector-type-blocks-update-populateslist-to-forlist

tests/schemas/invalid/test_invalid_relationship_list_doesnt_exist.json
```diff
@@ -0,0 +1,68 @@
...
+            "id": "relationship-question",
+            "type": "General",
+            "title": "Thinking of {first_person_name}, {second_person_name} is their <em>...</em>",
+            "answers": [{
+              "id": "relationship-answer",
+              "mandatory": true,
+              "type": "Relationship",
+              "playback": "{second_person_name} is {first_person_name} <em>…</em>",
+              "options": [{
+                  "value": "husband-or-wife",
```
ajmaddaford: The values should match the labels to be consistent with all the other schemas.

tests/schemas/invalid/test_invalid_relationship_list_doesnt_exist.json
```diff
@@ -0,0 +1,68 @@
...
+          "for_list": "not-a-list",
+          "question": {
+            "id": "relationship-question",
+            "type": "General",
+            "title": "Thinking of {first_person_name}, {second_person_name} is their <em>...</em>",
+            "answers": [{
+              "id": "relationship-answer",
+              "mandatory": true,
+              "type": "Relationship",
+              "playback": "{second_person_name} is {first_person_name} <em>…</em>",
```
bambrp: These would be the possessive form of first_person_name, though I appreciate the transform is not done here anyway. I suppose it could test for it? Or test for the right number of placeholders? Maybe that's a bit inflexible, enforcing something we wouldn't always do.

JamesGardiner: I think that's beyond the scope of the validator

ajmaddaford:
@JamesGardiner As discussed, this needs a check added that a `RelationshipCollector` only has one answer, and that answer type has to be of type `Relationship`.

PR #141, opened 2019-07-22
Add list collector summary definition to schema

JamesGardiner:
Adds a `summary` definition in the list collector json.

### DIFF
--- a/schemas/blocks/list_collector.json
+++ b/schemas/blocks/list_collector.json
@@ -18,6 +18,18 @@
     "required": ["id", "value"],
     "additionalProperties": false
   },
+  "summary": {
+    "type": "object",
+    "properties":{
+      "item_title": {
+          "$ref": "../string_interpolation/definitions.json#/string_with_placeholders"
+      }
+    },
+    "additionalProperties": false,
+    "required": [
+        "item_title"
+    ]
+  },
   "block": {
     "type": "object",
     "properties": {
@@ -51,6 +63,9 @@
       },
       "remove_block": {
         "$ref": "question.json#/block"
+      },
+      "summary": {
+        "$ref": "list_collector.json#/summary"
       }
     },
     "additionalProperties": false,
--- a/tests/schemas/valid/test_list_collector.json
+++ b/tests/schemas/valid/test_list_collector.json
@@ -200,6 +200,27 @@
               }]
             }]
           }
+        },
+        "summary": {
+          "item_title": {
+            "text": "{person_name}",
+            "placeholders": [{
+              "placeholder": "person_name",
+              "transforms": [{
+                "arguments": {
+                  "delimiter": " ",
+                  "list_to_concatenate": {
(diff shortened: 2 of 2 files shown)


JamesGardiner:
[Refactor list collector summary to be defined in schema](https://trello.com/c/8u0SzPws/3135-refactor-list-collector-summary-to-be-defined-in-schema)

schemas/blocks/list_collector.json
```diff
@@ -18,6 +18,35 @@
     "required": ["id", "value"],
     "additionalProperties": false
   },
+  "summary": {
+    "type": "object",
+    "properties":{
+      "type": {
+          "type": "string",
+          "enum": ["ListCollectorSummary"]
```
MebinAbraham: Slightly confused on why we have a type for a 'Summary' that is defined within the list collector itself?. What other types would there be?

schemas/blocks/list_collector.json
```diff
@@ -18,6 +18,35 @@
...
+              "description": "Summary text definitions using placeholders"
+          }
+        },
+        "required": [
+          "text",
+          "placeholders"
+        ]
+      }
+    },
+    "required": [
```
MebinAbraham: We should add `"additionalProperties": false`

schemas/blocks/list_collector.json
```diff
@@ -18,6 +18,35 @@
...
     "additionalProperties": false
   },
+  "summary": {
+    "type": "object",
+    "properties":{
+      "type": {
+          "type": "string",
+          "enum": ["ListCollectorSummary"]
+      },
+      "item_title": {
```
MebinAbraham: For `item_title` we should be able to reuse the `#/string_with_placeholders"` definition?

https://github.com/ONSdigital/eq-schema-validator/blob/refactor-list-collector-summary/schemas/string_interpolation/definitions.json#L3-L26

schemas/blocks/list_collector.json
JamesGardiner: You're right, this isn't necessary

schemas/blocks/list_collector.json
```diff
@@ -18,6 +18,18 @@
     "required": ["id", "value"],
     "additionalProperties": false
   },
+  "summary": {
```
MebinAbraham: Should we make this key `required`?

tests/schemas/valid/test_list_collector.json
```diff
@@ -200,6 +200,27 @@
               }]
             }]
           }
+        },
+        "summary": {
+          "item_titles": {
```
MebinAbraham: Should this not be singular?. An item has a title?

schemas/blocks/list_collector.json
JamesGardiner: Asked @ajmaddaford about this yesterday - the consensus was it shouldn't be mandatory

PR #143, opened 2019-08-01
Add Location Comparisons to when rules

iwootten:
### Motivation

Runner needs to be capable of displaying question variants based on the primary person, which is only available in the relationship location.

### Changes

Modify comparisons to be structured in objects coming from either a 'answers' or 'location' source.

    "comparison": {
      "source": "location",
      "id": "from_list_item_id"
    }

Additionally, I've modified RelationshipCollector validation to work with variants.

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -153,8 +153,12 @@ class Validator:  # pylint: disable=too-many-lines
                               'ListCollectors')
             elif block['type'] == 'RelationshipCollector':
                 errors.extend(self._validate_relationship_list_exists(block))
-                errors.extend(self._validate_relationship_collector_contains_single_relationship_answer(block))
-                errors.extend(self._validate_relationship_collector_answer_type(block))
+
+                if 'question_variants' in block:
+                    for variant in block['question_variants']:
+                        errors.extend(self._validate_relationship_collector_answers(variant['question']['answers']))
+                else:
+                    errors.extend(self._validate_relationship_collector_answers(block['question']['answers']))
 
             errors.extend(self._validate_questions(block, numeric_answer_ranges))
 
@@ -676,7 +680,7 @@ class Validator:  # pylint: disable=too-many-lines
     def _validate_when_rule(self, when_clause, answer_ids_with_group_id, referenced_id):
         """
         Validates any answer id in a when clause exists within the schema
-        Will also check that comparison_id exists
+        Will also check that comparison exists
         """
         errors = []
 
@@ -693,8 +697,8 @@ class Validator:  # pylint: disable=too-many-lines
             # We know the ids are correct, so can continue to perform validation
             errors.extend(self._validate_checkbox_exclusive_conditions_in_when_rule(when, answer_ids_with_group_id))
 
-            if 'comparison_id' in when:
-                errors.extend(self._validate_comparison_id_in_when_rule(when, answer_ids_with_group_id, referenced_id))
+            if 'comparison' in when:
+                errors.extend(self._validate_comparison_in_when_rule(when, answer_ids_with_group_id, referenced_id))
 
         return errors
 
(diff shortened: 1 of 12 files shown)


iwootten:
[Add primary person to relationships (M)](https://trello.com/c/B20tHhR5/3067-add-primary-person-to-relationships-m)

app/validation/validator.py
```diff
@@ -745,28 +755,29 @@ def _validate_checkbox_exclusive_conditions_in_when_rule(self, when, answer_ids_
...
         :return: list of dictionaries containing error messages, otherwise it returns an empty list
         """
         errors = []
 
-        answer_id, comparison_id, condition = when['id'], when['comparison_id'], when['condition']
-        comparison_answer_type = answer_ids_with_group_id[when['comparison_id']]['answer']['type']
-        id_answer_type = answer_ids_with_group_id[when['id']]['answer']['type']
-        conditions_requiring_list_match_values = ('equals any', 'not equals any', 'contains any', 'contains all')
+        if when['comparison']['source'] == 'answers':
+            answer_id, comparison_id, condition = when['id'], when['comparison']['id'], when['condition']
```
JamesGardiner: This would be more readable if the assigned variables were used in the following mapping accessors

app/validation/validator.py
```diff
@@ -153,8 +153,14 @@ def _validate_blocks(  # noqa: C901 pylint: disable=too-many-branches
                               'ListCollectors')
             elif block['type'] == 'RelationshipCollector':
                 errors.extend(self._validate_relationship_list_exists(block))
-                errors.extend(self._validate_relationship_collector_contains_single_relationship_answer(block))
-                errors.extend(self._validate_relationship_collector_answer_type(block))
+
+                if 'question_variants' in block:
```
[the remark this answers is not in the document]
iwootten: I'm happy to adopt the EAFP principle where it serves handling errors. But for this example neither route through can really be considered an exceptional condition whilst this suggestion arguably implies one of the routes is and (I think) makes the logical flow more difficult to follow. 

Given neither style has been officially adopted by the team and seems to be contentious within the Python community anyway I'm going to leave this as is.

app/validation/validator.py
```diff
@@ -705,13 +711,17 @@ def _validate_answer_ids_present_in_schema(self, when, answer_ids_with_group_id,
         :return: list of dictionaries containing error messages, otherwise it returns an empty list
         """
         errors = []
-        answer_reference_id_fields = ('id', 'comparison_id')
-        present_id_ref_keys = [x for x in answer_reference_id_fields if x in when]
+        ids_to_check = []
+
+        if 'id' in when:
```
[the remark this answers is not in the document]
iwootten: Following on from above, I don't think suppressing the errors in this way adds anything?

app/validation/validator.py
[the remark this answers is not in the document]
bitdivision: I would agree with Ian here, the if statements are simpler to  read for me here. The if statement also shows clearly which key is expected to be missing.

schemas/when_rule/definitions.json
```diff
@@ -41,9 +48,35 @@
...
+      "source": {
+        "type": "string",
+        "enum": ["answers"]
+      }
+    },
+    "required": ["id", "source"]
+  },
+  "location_comparison_object": {
+    "type": "object",
+    "description": "The object from which to obtain the value and compare to",
```
bitdivision: This should have a different description to answer_comparison_object.

app/validation/validator.py
```diff
@@ -153,8 +153,14 @@ def _validate_blocks(  # noqa: C901 pylint: disable=too-many-branches
                               'ListCollectors')
             elif block['type'] == 'RelationshipCollector':
                 errors.extend(self._validate_relationship_list_exists(block))
-                errors.extend(self._validate_relationship_collector_contains_single_relationship_answer(block))
-                errors.extend(self._validate_relationship_collector_answer_type(block))
+
+                if 'question_variants' in block:
+                    # Variants are checked to ensure they are consistent
+                    answers = block['question_variants'][0]['question']['answers']
```
bitdivision: It's probably worth checking all of the variants here. Unlikely to cause any major issues, but would save some time if you had a problem in your second variant.

bitdivision: I thought we had a method for getting all answers on a question, but I can't see it now.

app/validation/validator.py
```diff
@@ -745,28 +755,29 @@ def _validate_checkbox_exclusive_conditions_in_when_rule(self, when, answer_ids_
...
-        if condition in conditions_requiring_list_match_values:
-            if comparison_answer_type != 'Checkbox':
-                errors.append(self._error_message(
-                    f'The comparison_id `{comparison_id}` is not of answer type `Checkbox`. '
-                    f'The condition `{condition}` can only reference `Checkbox` answers when using `comparison id`'))
+            if condition in conditions_requiring_list_match_values:
+                if comparison_answer_type != 'Checkbox':
+                    errors.append(self._error_message(
+                        f'The comparison id `{comparison_id}` is not of answer type `Checkbox`. '
+                        f'The condition `{condition}` can only reference `Checkbox` answers when using `comparison id`'))
```
bitdivision: :+1:

schemas/when_rule/compare_to_location.json
```diff
@@ -0,0 +1,29 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "type": "object",
+  "description": "Checks if the location comparison is valid",
```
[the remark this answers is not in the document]
iwootten: Ah, ok - looks like this is also wrong on compare_to_answer_store then

bitdivision:
Needs a linting fix.

PR #144, opened 2019-08-06
Updated validator to allow values array in when conditions

SamGodwin2:
New functionality required in runner means we need to allow arrays of values to be passed into the when array in routing conditions, this Pr makes this change and allows it.


SamGodwin2:
[Spike - Change Runner to accept list of option ID (1 day)](https://trello.com/c/Z7TnIevV/926-spike-change-runner-to-accept-list-of-option-id-1-day)

PR #149, opened 2019-08-15
Allow detail_answer to be numeric

rmccar:
### PR Context

This PR is to allow a details_answers field to be numeric

This PR [#2279](ONSdigital/eq-survey-runner/pull/2279) needs this change to allow adding the validation needed

### Checklist

* [x] eq-translations updated to support any new schema keys which need translation


rmccar:
[Number write in field on radio option (on 'Other') (S)](https://trello.com/c/QwU5rLcl/3183-number-write-in-field-on-radio-option-on-other-s)

PR #150, opened 2019-08-15
Add pull request template

bitdivision:
### PR Context
Github only allows PR templates on the default branch. This checklist still makes sense for master though i think

### Checklist

* [ ] eq-translations updated to support any new schema keys which need translation


JamesGardiner:
Looks like codacy isn't a fan of your file name or markdown...

bitdivision:
Yeah, codacy is stupid. It matches runner. And isn't a problem

PR #151, opened 2019-08-16
Add ListCollector keys for SectionSummary

iwootten:
### PR Context

Adding new keys for summaries shown when a section contains list collectors. These are used when a ListCollector is shown on SectionSummary blocks.

### Checklist

* [x] eq-translations updated to support any new schema keys which need translation

### DIFF
--- a/schemas/blocks/list_collector.json
+++ b/schemas/blocks/list_collector.json
@@ -36,7 +36,7 @@
     },
     "additionalProperties": false,
     "required": [
-        "item_title", "add_link_text", "empty_list_text", "title"
+        "item_title", "title"
     ]
   },
   "block": {
--- a/schemas/blocks/list_collector.json
+++ b/schemas/blocks/list_collector.json
@@ -23,11 +23,20 @@
     "properties":{
       "item_title": {
           "$ref": "../string_interpolation/definitions.json#/string_with_placeholders"
+      },
+      "add_link_text": {
+        "type": "string"
+      },
+      "empty_list_text": {
+        "type": "string"
+      },
+      "title": {
+        "$ref": "../string_interpolation/definitions.json#/string_with_placeholders"
       }
     },
     "additionalProperties": false,
     "required": [
-        "item_title"
+        "item_title", "add_link_text", "empty_list_text", "title"
     ]
   },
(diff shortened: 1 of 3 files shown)


schemas/blocks/list_collector.json
```diff
@@ -23,11 +23,20 @@
...
+        "type": "string"
+      },
+      "title": {
+        "$ref": "../string_interpolation/definitions.json#/string_with_placeholders"
       }
     },
     "additionalProperties": false,
     "required": [
-        "item_title"
+        "item_title", "add_link_text", "empty_list_text", "title"
```
MebinAbraham: add_link_text, empty_list_text aren't actually required?. They are only needed for section summary, ie using list collector without a section summary?.
Unless the intention is that list collectors will always use a summary?

iwootten: I don't think it is, given that summary isn't actually required itself. I've removed them for now.

schemas/blocks/list_collector.json
```diff
@@ -66,6 +75,9 @@
       },
       "summary": {
         "$ref": "list_collector.json#/summary"
+      },
+      "hide_on_section_summary": {
```
ajmaddaford: This will also be needed on the primary person list collector, and possibly other blocks as there are some questions being added to the who lives here section in census that won't be displayed on the summary.

schemas/blocks/list_collector.json
```diff
@@ -23,11 +23,20 @@
     "properties":{
       "item_title": {
           "$ref": "../string_interpolation/definitions.json#/string_with_placeholders"
+      },
+      "add_link_text": {
+        "type": "string"
+      },
+      "empty_list_text": {
+        "type": "string"
```
ajmaddaford: These properties will need to be added to the eq-translations repo.

iwootten:
[Add lists to section summary (L)](https://trello.com/c/7ld7bDvO/3134-add-lists-to-section-summary-l)

schemas/blocks/list_collector.json
iwootten: As discussed offline, I've removed this now :)

PR #153, opened 2019-08-21
Add support for answer action

MebinAbraham:
### PR Context
Added support for answer actions feature introduced in https://github.com/ONSdigital/eq-survey-runner/pull/2301

### How to review
Test with https://github.com/ONSdigital/eq-survey-runner/pull/2301

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -238,17 +238,18 @@ class Validator:  # pylint: disable=too-many-lines
         for option in answer_options:
 
             action_params = option.get('action', {}).get('params')
-            if action_params:
+            if not action_params:
+                continue
 
-                list_name = action_params.get('list_name')
-                if list_name and list_name not in self._list_names:
-                    errors.append(
-                        (self._error_message(f'List name `{list_name}` defined in action params for answer `{answer["id"]}` does not exist')))
+            list_name = action_params.get('list_name')
+            if list_name and list_name not in self._list_names:
+                errors.append(
+                    (self._error_message(f'List name `{list_name}` defined in action params for answer `{answer["id"]}` does not exist')))
 
-                block_id = action_params.get('block_id')
-                if block_id and block_id not in self._block_ids:
-                    errors.append(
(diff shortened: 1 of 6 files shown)


MebinAbraham:
[Ability to route to a List Collector's add block.](https://trello.com/c/Yrl8kLIk/3221-ability-to-route-to-a-list-collectors-add-block)

app/validation/validator.py
```diff
@@ -229,6 +232,26 @@ def _validate_questions(self, block_or_variant, numeric_answer_ranges):
 
         return errors
 
+    def _validate_answer_actions(self, answer):
+        errors = []
+        answer_options = answer.get('options', {})
+        for option in answer_options:
+
+            action_params = option.get('action', {}).get('params')
```
```Python
try:
    action_params = option.get('action', {})['params']
except KeyError:
    continue
```

JamesGardiner: This will keep the code one level less nested

app/validation/validator.py
```diff
@@ -1434,6 +1457,19 @@ def _get_list_names(json_to_validate):
...
         return list_names
 
+    @staticmethod
+    def _get_block_ids(json_to_validate):
+        block_ids = []
+        for section in json_to_validate['sections']:
+            for group in section['groups']:
+                for block in group['blocks']:
+                    block_ids.append(block['id'])
+                    for sub_block in {'add_block', 'edit_block', 'remove_block'}:
```
ajmaddaford: This should also include `add_or_edit_block`

app/validation/validator.py
MebinAbraham: I rather keep it consistent as `if`s are already being used to check key existence.
If you really want me to reduce the indentation, I am happy to return early.

``` py
            action_params = option.get('action', {}).get('params')
            if not action_params:
                continue

            list_name = action_params.get('list_name')
```

app/validation/validator.py
MebinAbraham: Ah ofc. Added.

app/validation/validator.py
JamesGardiner: Yep, good compromise

PR #164, opened 2019-10-04
Allow lists to be displayed on questions

iwootten:
### PR Context

Add a list_summary key to questions to allow lists to be displayed on questions.


schemas/blocks/question.json
```diff
@@ -22,6 +22,26 @@
       "question_variants": {
         "$ref": "definitions.json#/question_variants"
       },
+      "list_summary": {
```
MebinAbraham: Should we add this to content blocks as well?

tests/schemas/valid/test_list_on_question.json
```diff
@@ -0,0 +1,351 @@
+{
```
MebinAbraham: Can we name this `test_list_summary_on_question.json` for clarity?

schemas/blocks/question.json
iwootten: I don't know, but we don't have a requirement for this right now. We'd need to make changes to display it in the open PR on runner, which would require checking for the key in the content handler there. Seems overkill for now.

PR #166, opened 2019-10-09
Add quote validation

berroar:
### PR Context
Added validation to check for the use of dumb quotes within lines of text. We should only be using smart quotes within our schema text. See http://smartquotesforsmartpeople.com/

### How to Review
Check the new tests created that find dumb quotes.
Run the validation tests against the new test schema.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,8 @@
-FROM python:3.6-alpine
+FROM python:3.6-slim
+
+RUN apt-get update && apt-get install --no-install-recommends -y git=1:2.20.1-2 \
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/*
 
 RUN pip install pipenv==2018.11.26
 
--- a/Pipfile
+++ b/Pipfile
@@ -8,10 +8,11 @@ name = "pypi"
 [packages]
 
 Flask = "*"
-jsonschema = "*"
+jsonschema = "<3"
 structlog = "*"
 python-dateutil = "*"
-
+jsonpointer = "*"
+eq-translations = {git = "https://github.com/ONSDigital/eq-translations.git",ref = "master"}
 
 [dev-packages]
 
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -5,7 +5,10 @@ from collections import defaultdict
 from datetime import datetime
 from json import load
 
+
 from dateutil.relativedelta import relativedelta
+from eq_translations.survey_schema import SurveySchema
+from jsonpointer import resolve_pointer
 from jsonschema import SchemaError, RefResolver, validate, ValidationError
 from jsonschema.exceptions import best_match
 
@@ -36,6 +39,7 @@ class Validator:  # pylint: disable=too-many-lines
         validation_errors = []
         validation_errors.extend(self._validate_schema_contain_metadata(json_to_validate))
         validation_errors.extend(self._validate_duplicates(json_to_validate))
+        validation_errors.extend(self._validate_smart_quotes(json_to_validate))
 
         section_ids = []
 
@@ -1358,6 +1362,22 @@ class Validator:  # pylint: disable=too-many-lines
                 errors.append(self._error_message(
                     "`previous_transform` not referenced in chained transform in block id '{}'".format(
                         block_id)))
+
+        return errors
+
+    def _validate_smart_quotes(self, json_schema):
+
+        schema_object = SurveySchema()
+        schema_object.schema = json_schema
+        errors = []
+
+        # pylint: disable=invalid-string-quote
+        quote_regex = re.compile(r"['|\"]+(?![^{]*})+(?![^<]*>)")
+
+        for pointer in schema_object.pointers:
+            schema_text = resolve_pointer(json_schema, pointer)
+            if quote_regex.search(schema_text):
+                errors.append(self._error_message(f'Found dumb quotes(s) in schema text at {pointer}'))
 
         return errors
 
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "a41563ee802f74c405deb68e598cf533b0e4aca3418fa149f2dda1f4fd30daa1"
+            "sha256": "681207747e17bbd5da6b44b6d54db9d85f5642e003a2adca6b3ffa9736b576e2"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -23,13 +23,17 @@
             ],
             "version": "==7.0"
         },
+        "eq-translations": {
+            "git": "https://github.com/ONSDigital/eq-translations.git",
+            "ref": "a7fbf8ca1b3395c99e37226334ab51baea63bb8e"
+        },
         "flask": {
             "hashes": [
-                "sha256:2271c0070dbcb5275fad4a82e29f23ab92682dc45f9dfbc22c02ba9b9322ce48",
-                "sha256:a080b744b7e345ccfcbc77954861cb05b3c63786e93f2b3875e0913d44b43f05"
+                "sha256:13f9f196f330c7c2c5d7a5cf91af894110ca0215ac051b5844701f2bfd934d52",
+                "sha256:45eb5a6fd193d6cf7e0cf5d8a5b31f83d5faae0293695626f539a823e93b13f6"
             ],
             "index": "pypi",
-            "version": "==1.0.2"
+            "version": "==1.1.1"
         },
         "itsdangerous": {
             "hashes": [
@@ -40,10 +44,18 @@
         },
         "jinja2": {
             "hashes": [
-                "sha256:74c935a1b8bb9a3947c50a54766a969d4846290e1e788ea44c1392163723c3bd",
-                "sha256:f84be1bb0040caca4cea721fcbbbbd61f9be9464ca236387158b0feea01914a4"
+                "sha256:74320bb91f31270f9551d46522e33af46a80c3d619f4a4bf42b3164d30b5911f",
+                "sha256:9fe95f19286cfefaa917656583d020be14e7859c6b0252588391e47db34527de"
             ],
-            "version": "==2.10"
+            "version": "==2.10.3"
+        },
+        "jsonpointer": {
+            "hashes": [
+                "sha256:c192ba86648e05fdae4f08a17ec25180a9aef5008d973407b581798a83975362",
+                "sha256:ff379fa021d1b81ab539f5ec467c7745beb1a5671463f9dcc2b2d458bd361c1e"
+            ],
+            "index": "pypi",
+            "version": "==2.0"
         },
         "jsonschema": {
             "hashes": [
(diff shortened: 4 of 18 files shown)


berroar:
[Inconsistent quotes used in Welsh translations](https://trello.com/c/6LEeu41c/3228-inconsistent-quotes-used-in-welsh-translations)

app/validation/validator.py
```diff
@@ -1361,6 +1363,93 @@ def _validate_placeholder_transforms(self, transforms, block_id):
...
+                    'Found dumb quotes in question/answer guidance for id {}'.format(sub_block_id)))
+            if ('description' in sub_block_contents[0]) and \
+                    (self._search_dumb_quotes(sub_block_contents[0]['description'])):
+                errors.append(self._error_message(
+                    'Found dumb quotes in question/answer guidance for id {}'.format(sub_block_id)))
+            if 'list' in sub_block_contents[0]:
+                for item in sub_block_contents[0]['list']:
+                    if self._search_dumb_quotes(str(item)):
+                        errors.append(self._error_message(
+                            'Found dumb quotes in question/answer guidance for id {}'.format(sub_block_id)))
```
ajmaddaford: In `eq-translations` we search for all text elements in a schema (have a look at the `no_context_pointers` and `context_pointers` methods). We should look to extract that into something common that we can use here.

app/validation/validator.py
```diff
@@ -1361,6 +1363,93 @@ def _validate_placeholder_transforms(self, transforms, block_id):
...
+            question_title = question['title']['text']
+
+        if self._search_dumb_quotes(question_title) is not None:
+            errors.append(self._error_message(
+                'Found dumb quotes in question title for question id {}'.format(question['id'])))
+
+        if 'description' in question:
+            if self._search_dumb_quotes(question['description']):
+                errors.append(self._error_message(
+                    'Found dumb quotes in question description for question id {}'.format(question['id'])))
```
```python
try:
    msg = f'Found dumb quotes in question description for question id {question_id}'
    description_error = self._error_message(msg) if self._search_dumb_quotes(question['description']) else None
except KeyError:
    pass

File: app/validation/validator.py (hunk 1 above)
Comment (berroar) [reply app/validation/validator.py]: Refactored with code from eq-translations

File: app/validation/validator.py [hunk 3]
```diff
@@ -1361,6 +1363,93 @@ def _validate_placeholder_transforms(self, transforms, block_id):
 
         return errors
 
+    def _validate_dumb_quotes_in_questions(self, question):
+        errors = []
+
+        if isinstance(question['title'], str):
+            question_title = question['title']
+        else:
+            question_title = question['title']['text']
```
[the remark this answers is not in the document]
berroar: Changed the solution, see Andrew's comment above.

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,101 @@
+import re
+from jsonpointer import resolve_pointer
+
+
+def get_pointer_keys():
```
iwootten: I'm not sure duplicating some of the methods of the translations repo is the best approach here.

Given these are likely to change over time - it probably makes sense that eq-translations is installed as dependency here and then the relevant utility methods within it are used to locate pointers of interest. Much like our runner does to translate schemas at the moment.

berroar: Have refactored and added eq-translations as a dependency

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,64 @@
...
+    list_pointers = schema_object.get_list_pointers()
+    list_contents = get_list_pointer_contents(json_schema, list_pointers)
+
+    for list_text in list_contents:
+        if search_dumb_quotes(list_text):
+            return error_message
+
+    for schema_text in schema_contents:
+        if search_dumb_quotes(schema_text):
+            return error_message
```
ajmaddaford: - We should be generating an error for each string that contains a dumb quote. 
- We should create a single consolidated list of pointers, and then iterate over that.

Something like:

```python
pointers = []
errors = []

for key in get_pointer_keys():
    pointers.extend(find_pointers_to(json_schema, key))

schema_object = SurveySchema()
schema_object.load(schema_path)
pointers.extend(schema_object.get_list_pointers())

for pointer in pointers:
    schema_text = resolve_pointer(schema, pointer)
    if search_dumb_quotes(list_text):
        errors.append(f'Found dumb quote(s) in schema text at [{pointer}]')
```

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,64 @@
...
+def search_dumb_quotes(text):
+    return re.search(r"'", text)
+
+
+def validate_json_schema_text(json_schema, schema_path):
+    error_message = 'Found dumb quotes in schema text'
+    schema_contents = get_pointer_contents(json_schema, get_pointer_keys())
+
+    schema_object = SurveySchema()
+    schema_object.load(schema_path)
```
ajmaddaford: `schema_path` may not be needed, we may be able to do:

```python
schema_object = SurveySchema()
schema_object.schema = json_schema
```

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,64 @@
...
+def get_list_pointer_contents(schema, schema_elements):
+    contents = []
+    for element in schema_elements:
+        pointer_contents = resolve_pointer(schema, element)
+        contents.append(pointer_contents)
+    return contents
+
+
+def search_dumb_quotes(text):
+    return re.search(r"'", text)
```
ajmaddaford: If the other changes are made and this is only called from one place, I don't think this should be a separate method.

app/validation/validate_schema_text.py
berroar: Refactored this

app/validation/validate_schema_text.py
berroar: Fixed

app/validation/validate_schema_text.py
berroar: Removed the method

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,49 @@
...
+        'add_link_text',
+        'empty_list_text',
+        'text',
+        'title',
+    ]
+
+    return pointer_keys
+
+
+def get_list_pointer_contents(schema, schema_elements):
```
ajmaddaford: This method is no longer used.

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,49 @@
...
+    for key in get_pointer_keys():
+        pointers.extend(find_pointers_to(json_schema, key))
+
+    schema_object = SurveySchema()
+    schema_object.schema = json_schema
+    pointers.extend(schema_object.get_list_pointers())
+
+    for pointer in pointers:
+        schema_text = resolve_pointer(json_schema, pointer)
+        if re.search(r"'", schema_text):
```
ajmaddaford: The regex should be compiled before the for loop, so it's not done on every iteration:

```python
quote_regex = re.compile(r"'")
for pointer in pointers:
    schema_text = resolve_pointer(json_schema, pointer)
    if quote_regex.search(schema_text):
```

Also, the regex should also search for `"`.

berroar: Fixed this

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,43 @@
+import re
```
JamesGardiner: Theses imports need to be split to follow PEP8 standards

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,43 @@
...
+from jsonpointer import resolve_pointer
+from eq_translations.utils import find_pointers_to
+from eq_translations.survey_schema import SurveySchema
+
+
+def get_pointer_keys():
```
JamesGardiner:
```suggestion
```

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,43 @@
...
+        'playback',
+        'item_title',
+        'label',
+        'add_link_text',
+        'empty_list_text',
+        'text',
+        'title',
+    ]
+
+    return pointer_keys
```
JamesGardiner:
```suggestion
```

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def validate_json_schema_text(json_schema):
+    pointers = []
+    errors = []
+
+    for key in get_pointer_keys():
```
JamesGardiner:
```suggestion
    pointers = [find_pointers_to(json_schema, key) for key in pointer_keys] + schema_object.get_list_pointers()
```

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,43 @@
...
+        'empty_list_text',
+        'text',
+        'title',
+    ]
+
+    return pointer_keys
+
+
+def validate_json_schema_text(json_schema):
+    pointers = []
```
JamesGardiner:
```suggestion
```

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,43 @@
...
+    errors = []
```
JamesGardiner:
```suggestion
```

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,43 @@
...
+
+    for key in get_pointer_keys():
+        pointers.extend(find_pointers_to(json_schema, key))
```
JamesGardiner:
```suggestion
```

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,43 @@
...
+
+    schema_object = SurveySchema()
```
JamesGardiner: These can be moved to the start of the function

app/validation/validate_schema_text.py
```diff
@@ -0,0 +1,43 @@
...
+    schema_object.schema = json_schema
+    pointers.extend(schema_object.get_list_pointers())
```
JamesGardiner:
```suggestion
```

tests/test_schema_text.py
```diff
@@ -0,0 +1,50 @@
...
+
+from structlog import configure
+from structlog import getLogger
+from structlog.stdlib import LoggerFactory
+
+from app.validation.validate_schema_text import validate_json_schema_text
+
+logger = getLogger()
+
+configure(logger_factory=LoggerFactory())
```
JamesGardiner: Is the logger configuration needed? I don't see any log lines in the below code

tests/test_schema_text.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def validate_schema_text(json_to_validate):
+    errors = validate_json_schema_text(json_to_validate)
+    return errors
+
+
+def _open_and_load_schema_file(file):
+    json_file = open(os.path.join(os.path.dirname(os.path.realpath(__file__)), file), encoding='utf8')
```
JamesGardiner: use a context manager to open a file

tests/test_schema_text.py
```diff
@@ -0,0 +1,50 @@
...
+    json_to_validate = load(json_file)
+
+    return json_to_validate
```
JamesGardiner: There isn't any point assigning the opened file - just return from within a context manager. The `__exit__` method will always be called and close the file handler correctly

tests/test_schema_text.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def test_param_valid_schemas(valid_schema_filename):
+    """
+    Uses py.test generated tests to validate all schemas contained in the 'valid' folder.
+    """
+    with open(valid_schema_filename, encoding='utf8') as json_file:
+        json_to_validate = load(json_file)
```
JamesGardiner: Any reason you aren't using the _open_and_load_schema_file function here

app/validation/validate_schema_text.py
berroar: Added this to the main function.

tests/test_schema_text.py
berroar: I have removed these

tests/test_schema_text.py
berroar: Amended the open and load function to use a context manager

tests/test_schema_text.py
berroar: As above

app/validation/validate_schema_text.py
berroar: I don't think I am able to make this change as I need the return to be a flattened list of strings.

berroar: No longer required after latest refactor

app/validation/validate_schema_text.py
berroar: Outdated

app/validation/validate_schema_text.py
berroar: Outdated

app/validation/validate_schema_text.py
berroar: Outdated

Pipfile
```diff
@@ -8,10 +8,11 @@ name = "pypi"
 [packages]
 
 Flask = "*"
-jsonschema = "*"
+jsonschema = "<3"
 structlog = "*"
 python-dateutil = "*"
-
+jsonpointer = "*"
+eq-translations = {editable = true,git = "https://github.com/ONSDigital/eq-translations.git",ref = "master"}
```
JamesGardiner: Does eq-translations need to be installed with the editable flag? This has the effect of adding all the eq-translation dependencies to the Pipfile.lock

berroar: Removed this

app/validation/validator.py
```diff
@@ -1361,6 +1365,22 @@ def _validate_placeholder_transforms(self, transforms, block_id):
...
+        schema_object.schema = json_schema
+        errors = []
+
+        # pylint: disable=invalid-string-quote
+        quote_regex = re.compile(r"['|\"]+(?![^{]*})+(?![^<]*>)")
+
+        for pointer in schema_object.pointers:
+            schema_text = resolve_pointer(json_schema, pointer)
+            if quote_regex.search(schema_text):
+                errors.append(self._error_message(f'Found dumb quotes(s) in schema text at [{pointer}]'))
```
iwootten: The square braces are a bit confusing here - they result in error messages that appear to be indicating lists.

berroar: I will remove these.

Dockerfile
```diff
@@ -1,4 +1,8 @@
-FROM python:3.6-alpine
+FROM python:3.6-slim
```
JamesGardiner: Does bumping to 3.8 break anything?

app/validation/validator.py
```diff
@@ -5,6 +5,9 @@
 from datetime import datetime
 from json import load
 
+
+from jsonpointer import resolve_pointer
```
JamesGardiner: Blocks should be alphabetical

Dockerfile
iwootten: The Pipfile is based on 3.6, so presumably it wouldn't be installed?

JamesGardiner: Sorry @berroar I should have marked this as a comment not a request for change.

JamesGardiner: Sorry @berroar I should have marked this as a comment not a request for change.

app/validation/validator.py
berroar: Fixed

berroar:
Need to wait for https://github.com/ONSdigital/eq-survey-runner/pull/2486 before merging.

PR #172, opened 2019-11-13
Add rule to enforce a direct match between answer labels and values

berroar:
### PR Context
We have found instances in a number of schemas where answer labels and values do not match. See [trello](https://trello.com/c/rXoPEdks/3367-review-schemas-for-value-and-labels-mismatching)

We need to write a validator rule to enforce the direct match - the answer value should match the label.

### How to review
Check the new validation rule that has been added to ensure that answer labels and values match. Check that all tests pass.


app/validation/validator.py
```diff
@@ -281,6 +282,16 @@ def _validate_answer_actions(self, answer):
...
 
+    def _ensure_answer_labels_and_values_match(self, answer):
+        errors = []
+        if 'options' in answer:
+            for options in answer['options']:
+                if 'placeholders' not in options['label']:
+                    if options['label'] != options['value']:
+                        errors.append(
+                            (self._error_message(f'`Found mismatching answer value(s) and label(s) for {answer["id"]}`')))
+        return errors
```
MebinAbraham: Worth reducing some nesting. We also seem to already use the pattern of `isinstance` for option labels.
```py
    def _ensure_answer_labels_and_values_match(self, answer):
        errors = []
        for option in answer.get('options', []):
            if isinstance(option['label'], str) and option['label'] != option['value']:
                errors.append(
                    (self._error_message(f'`Found mismatching answer value(s) and label(s) for {answer["id"]}`')))

        return errors
```

berroar: Fixed this

berroar:
[Review schemas for value and labels mismatching](https://trello.com/c/rXoPEdks/3367-review-schemas-for-value-and-labels-mismatching)

tests/schemas/invalid/test_invalid_mismatching_answer_label_and_value.json
```diff
@@ -0,0 +1,207 @@
+{
```
MebinAbraham: Can we reduce this schema and get rid of all the things we don't need like routing rules etc. All we need is a 1 question with answer options that has a mismatch. Worth adding a placeholder one too.

MebinAbraham:
We should review and discuss the use of dynamic option value, ie where they do differ and whether there should be a pattern for it.

tests/schemas/invalid/test_invalid_mismatching_answer_label_and_value.json
berroar: I've simplified the schema as best as I could with the addition of a placeholder question.

app/validation/validator.py
```diff
@@ -281,6 +282,15 @@ def _validate_answer_actions(self, answer):
 
         return errors
 
+    def _ensure_answer_labels_and_values_match(self, answer):
+        errors = []
+        for option in answer.get('options', []):
+            if isinstance(option['label'], str) and option['label'] != option['value']:
+                errors.append(
+                    (self._error_message(f'`Found mismatching answer value(s) and label(s) for {answer["id"]}`')))
```
ajmaddaford: We also have the answer label here, so it would be useful to include that in the error message, and then we can also then drop the `(s)`'s

tests/schemas/invalid/test_invalid_mismatching_answer_label_and_value.json
```diff
@@ -0,0 +1,130 @@
...
+                    "q_code": "1",
+                    "type": "TextField"
+                  },
+                  {
+                    "id": "surname-answer",
+                    "label": "What is your surname",
+                    "mandatory": false,
+                    "q_code": "2",
+                    "type": "TextField"
+                  }
```
ajmaddaford: This schema could be simplified further:
- Single name field in first block
- Drop `q_code` properties
- Can the `tea-or-coffee` block be dropped and the `No` answer label for the `correct` block be changed to not match the label?

app/validation/validator.py
berroar: Updated this

tests/schemas/invalid/test_invalid_mismatching_answer_label_and_value.json
berroar: Have changed the validator so that it checks for dynamic labels and have updated the schema.

app/validation/validator.py
```diff
@@ -281,6 +282,21 @@ def _validate_answer_actions(self, answer):
...
+            if isinstance(option['label'], str):
+                if option['label'] != option['value']:
+                    errors.append(
+                        (self._error_message(f'`Found mismatching answer value for label: {option["label"]} '
+                                             f'in answer id: {answer["id"]}`')))
+            else:
+                if option['label']['text'] != option['value']:
+                    errors.append(
+                        (self._error_message(f'`Found mismatching answer value for label: {option["label"]["text"]} '
+                                             f'in answer id: {answer["id"]}`')))
```
ajmaddaford: This could be simplified to:

```python
if isinstance(option['label'], str):
    label = option['label']
else:
    label = option['label']['text']
           
if label != option['value']:
    errors.append(
        (self._error_message(f'`Found mismatching answer value for label: {label} '
                             f'in answer id: {answer["id"]}`')))
```

berroar: Refactored this

PR #174, opened 2019-11-15
Add method to validate answer values used in when rules exist

petechd:
This adds a method of checking if answer values used in routing exist. It deliberately ignores numbers only being used in routing(in test schemas) because answer's form number fields don't have answer 'value' set by default(so validator's method would always fail). [Fix mismatching labels and values](https://github.com/ONSdigital/eq-questionnaire-runner/pull/3) needs to be merged to make new method ignore any census schemas other than English ones. There are 4 routing errors in eq-questionnaire-runner's English language and test schemas which won't pass this validator check at the moment, they need to be fixed before merging this PR.[Trello card](https://trello.com/c/EbHSxOaI/3481-validate-answer-values-used-in-routing-exist)

### How to review 
Use a local instance of validator to test schemas and run this script in local copy of runner: 
`./scripts/test_schemas.sh --local`
All missing answer values in routing of all schemas should be detected.


petechd:
[Validate answer values used in routing exist](https://trello.com/c/EbHSxOaI/3481-validate-answer-values-used-in-routing-exist)

app/validation/validator.py
```diff
@@ -506,6 +507,59 @@ def _validate_routing_rule(self, rule, answer_ids_with_group_id, block_or_group)
...
+                for block in answer_ids_with_parent_id:
+                    if 'options' in answer_ids_with_parent_id[block]['answer']:
+                        for answer_block in answer_ids_with_parent_id[block]['answer'].get('options'):
+                            if rule_value == answer_block.get('value'):
+                                amount_matched += 1
+                if amount_matched == 0:
+                    errors.append(Validator._error_message(
+                        f'Answer option and routing rule values mismatch, missing answer value: '
+                        f'{rule_value}'))
+        return errors
```
JamesGardiner: This needs refactoring - there is scope to make it significantly less complex.

1. You could make a generic function that takes a value, then pass each item from a `values` iterable as well as single `value` items to it, rather than having two very similar functions.
1. A  few of the if statements would be unnecessary if default values were used in the `.get` methods
1. There is a mixture of f-string and old style `.format` - f-strings are preferred
1. Do these need to be static methods?
1. There is a lot of nesting - a flat strucutre is generally better in terms of readability - is there a way to flatten this down?

petechd: I think it's all addressed now, to a degree(except @staticmethod)

app/validation/validator.py
```diff
@@ -506,6 +507,45 @@ def _validate_routing_rule(self, rule, answer_ids_with_group_id, block_or_group)
...
+            for block in answer_ids_with_parent_id:
+                for answer_block in answer_ids_with_parent_id[block]['answer'].get('options', {}):
+                    if rule_value == answer_block.get('value'):
+                        amount_matched += 1
+            if amount_matched == 0:
+                errors.append(Validator._error_message(
+                    f'Answer option and routing rule values mismatch, missing answer value: '
+                    f'{rule_value}'))
+        return errors
+
```
iwootten: This should all form part of the existing "validate_routing_rule" method, as that's what it is doing. I think a better solution may be to create a map of answer ids to possible option values and then use this during validation.

iwootten: This could be reduced to a method which determines if a single rule value is valid and returns it as a boolean.

app/validation/validator.py
```diff
@@ -506,6 +507,45 @@ def _validate_routing_rule(self, rule, answer_ids_with_group_id, block_or_group)
...
+        if when_dict.get('value'):
+            rule_value = when_dict.get('value')
+            errors.extend(self._validate_routing_when_dict_value(answer_ids_with_parent_id, amount_matched, rule_value))
+
+        elif when_dict.get('values'):
+            for rule_value in when_dict.get('values'):
+                errors.extend(self._validate_routing_when_dict_value(
+                    answer_ids_with_parent_id, amount_matched, rule_value))
+
+        return errors
```
iwootten: Both these calls are doing the same thing, once for values and another for value. You can save the nesting here if you make both routes use the same call.

```python
errors = []
when_values = when_dict.get('values', [])

if not when_values:
    when_values.append(when_dict.get('value'))

for rule_value in when_values:
    valid_rule = self.is_rule_value_valid(
                    answer_ids_with_parent_id, rule_value))

   if not valid_rule:
      errors.append(f'Answer option and routing rule values mismatch, missing answer value: {rule_value}')

File: app/validation/validator.py [hunk 4]
```diff
@@ -1491,11 +1520,19 @@ def _get_answers_with_parent_ids(self, json_to_validate):
...
+                    answers[answer['id']] = {
+                        'answer': answer,
+                        **context
+                    }
+
                 for option in answer.get('options', []):
+                    answer_options = answers[answer['id']]['answer']['options']
+
+                    if answer_options and option['value'] not in [option['value'] for option in answer_options]:
+                        answer_options.append(option)
```
iwootten: These answer_options don't appear to be used for anything other than these three lines?

app/validation/validator.py
```diff
@@ -1491,11 +1520,19 @@ def _get_answers_with_parent_ids(self, json_to_validate):
         answers = {}
         for question, context in self._get_questions_with_context(json_to_validate):
             for answer in question.get('answers', []):
-                answers[answer['id']] = {
-                    'answer': answer,
-                    **context
-                }
+
+                if not answer['id'] in answers:
```
iwootten: I think this check may be redundant? Won't the context will be the same in cases where the answer id is the same?

app/validation/validator.py
```diff
@@ -518,9 +518,38 @@ def _validate_routing_rule(self, rule, answer_ids_with_group_id, block_or_group)
...
+
+            if not valid_rule:
+                errors.append(Validator._error_message(f'Answer option and routing rule values mismatch, '
+                                                       f'missing answer value: {rule_value}'))
+
+        return errors
+
+    @staticmethod
+    def is_rule_value_valid(answer_ids_with_parent_id, rule_value):
+        if isinstance(rule_value, int) is not True:
```
iwootten: You can save nesting here if you return early.

```suggestion
        if isinstance(rule_value, int) is True:
           return True
        for block in answer_ids_with_parent_id:
           ...
           ...
```

iwootten: Additionally, I think have unit tests for these methods added will help.

petechd: I test it already in test_invalid_routing_answer_mismatch.json, `{
                  "goto": {
                    "block": "country-interstitial-one",
                    "when": [
                      {
                        "id": "country-checkbox-answer",
                        "condition": "contains",
                        "value": 7
                      }
                    ]
                  }
                },`

app/validation/validator.py
```diff
@@ -542,13 +542,12 @@ def _validate_routing_when_dict(self, when_dict, answer_ids_with_parent_id):
...
-                    if rule_value == answer_block.get('value'):
-                        return True
-            return False
-        return True
+        if isinstance(rule_value, int) is True:
+            return True
+        for block in answer_ids_with_parent_id:
+            for answer_block in answer_ids_with_parent_id[block]['answer'].get('options', {}):
+                if rule_value == answer_block.get('value'):
+                    return True
```
iwootten: This appears to never return false.

app/validation/validator.py
petechd: It checks for all possible variants of proxy and non proxy answers(which have same context but different values and labels, paired with different routing rules). If you remove that only answers from last variant will be added and validator will fail.

app/validation/validator.py
iwootten: I don't think that's the best example for testing this single method, we should be explicitly adding tests for each bit of behaviour we introduce.

app/validation/validator.py
```diff
@@ -518,9 +518,38 @@ def _validate_routing_rule(self, rule, answer_ids_with_group_id, block_or_group)
...
+
+            if not valid_rule:
+                errors.append(Validator._error_message(f'Answer option and routing rule values mismatch, '
+                                                       f'missing answer value: {rule_value}'))
+
+        return errors
+
+    @staticmethod
+    def is_rule_value_valid(answer_ids_with_parent_id, rule_value):
+        if isinstance(rule_value, int) is True:
```
```suggestion
        if isinstance(rule_value, int):
```

app/validation/validator.py
iwootten: Won't the current approach allow invalid answers from alternative variants then?

petechd: Removing answer variants from pool of answers will remove some of values referenced by routing values. Routing rules are sometimes looking for specific variants of same questions, but if we add all variants of given question, we will end up with same labels and values which will break validation. So it's not all variants, it's only all different variants. Routing values are searching for answer values not the other way round. There was a bug in answer_ids_with group_ids creation when last variant of question was overwriting the previous variant, so even if you had four, you ended up with last.

tests/test_rule_validation.py
```diff
@@ -0,0 +1,125 @@
...
+        comparison = validation.Validator().is_rule_value_valid(
+            single_answer_id, "No I need to change this"
+        )
+
+        assert comparison is True
+
+    @staticmethod
+    def test_number_rule_ignored():
+
+        comparison = validation.Validator().is_rule_value_valid(single_answer_id, 123)
```
iwootten: Shouldn't this return false? The value isn't possible to enter with the rule given. The test also seems to be inappropriately named, as the return value implies it hasn't been ignored.

tests/test_rule_validation.py
```diff
@@ -0,0 +1,125 @@
...
+                    "mandatory": False,
+                    "type": "Number",
+                    "default": 0,
+                },
+                "block": "number-of-employees-total-block",
+                "group_id": "confirmation-block",
+                "section": "default-section",
+            }
+        }
+        comparison = validation.Validator().is_rule_value_valid(answer_id, "Ignored")
```
iwootten: You could do: 
```suggestion
        from app.validation.validator import Validator
        ....
        comparison = Validator.is_rule_value_valid(answer_id, "Ignored")
```

tests/test_rule_validation.py
petechd: If this returned false it would not find an answer and continue searching through answers map. Integer routing values are ignored because tracking them in answers map is problematic(they don't necessarily need to exist in answer values to work as routing rules).

iwootten: This role of this method isn't to locate an answer, it's to test if a given answer is valid against the rule that defines it. If you need to allow other values, you should add further methods to allow for this.

At the moment this reads as the method confirms is_rule_valid(some_map_of_valid_values, invalid_value), which isn't correct.

petechd: But in some instances, integer rule value cannot be valid against any answer in schema, yet still be valid. Validator isn't able to check that because such answer value doesn't exist. This test checks if integer rule is being ignored as expected.

app/validation/validator.py
```diff
@@ -729,9 +729,49 @@ def _validate_routing_rule(self, rule, answer_ids_with_group_id, block_or_group)
...
+                        f"missing answer value: {rule_value}"
+                    )
+                )
+
+        return errors
+
+    @staticmethod
+    def is_rule_value_valid(answer_ids_with_parent_id, rule_value):
+        if isinstance(rule_value, int):
+            return True
```
iwootten: I think this check would be better placed in the parent method.

petechd: Placed it in the parent now. This changed the numbers test to assert "false" now.

tests/test_rule_validation.py
```diff
@@ -0,0 +1,125 @@
...
+    def test_search_first_value_in_options():
+
+        comparison = validation.Validator().is_rule_value_valid(
+            single_answer_id, "Yes this is correct"
+        )
+
+        assert comparison is True
+
+    @staticmethod
+    def test_search_second_value_in_options():
```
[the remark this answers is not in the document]
petechd: Done!

tests/test_rule_validation.py
```diff
@@ -0,0 +1,125 @@
...
+            "mandatory": True,
+            "q_code": "d50",
+        },
+        "block": "confirm-zero-employees-block",
+        "group_id": "confirmation-block",
+        "section": "default-section",
+    }
+}
+
+multiple_answer_ids = {
```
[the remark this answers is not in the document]
petechd: Changed!

tests/test_rule_validation.py
```diff
@@ -0,0 +1,125 @@
+import unittest
+
+import app.validation.validator as validation
+
+single_answer_id = {
```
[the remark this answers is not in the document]
petechd: Done!

app/validation/validator.py
petechd: That's done, thanks!

app/validation/validator.py
petechd: Changed it, thanks!

app/validation/validator.py
petechd: Removed!

tests/test_rule_validation.py
petechd: This is addressed now, thanks!

app/validation/validator.py
```diff
@@ -729,9 +729,49 @@ def _validate_routing_rule(self, rule, answer_ids_with_group_id, block_or_group)
...
+        for check_block in answer_ids_with_parent_id:
+            if answer_ids_with_parent_id[check_block]["answer"].get("options"):
+                for answer_block in answer_ids_with_parent_id[check_block][
```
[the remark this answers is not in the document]
petechd: Solved, thanks!

app/validation/validator.py
```diff
@@ -729,9 +729,49 @@ def _validate_routing_rule(self, rule, answer_ids_with_group_id, block_or_group)
...
+                )
+
+        return errors
+
+    @staticmethod
+    def is_rule_value_valid(answer_ids_with_parent_id, rule_value):
+        if isinstance(rule_value, int):
+            return True
+        for check_block in answer_ids_with_parent_id:
+            if answer_ids_with_parent_id[check_block]["answer"].get("options"):
```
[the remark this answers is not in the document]
petechd: Solved, thanks!

app/validation/validator.py
```diff
@@ -742,16 +742,23 @@ def _validate_routing_when_dict(self, when_dict, answer_ids_with_parent_id):
...
         if when_value:
             when_values.append(when_value)
 
         for rule_value in when_values:
-            if isinstance(rule_value, int):
-                valid_rule = True
-            else:
+            valid_rule = False
+
+            if answer_id is None:
```
iwootten: This could be moved above this loop.

app/validation/validator.py
```diff
@@ -742,16 +742,23 @@ def _validate_routing_when_dict(self, when_dict, answer_ids_with_parent_id):
...
+                return errors
+
+            if answer_ids_with_parent_id[answer_id]["answer"].get("type") in (
+                "Radio",
+                "Checkbox",
+            ):
                 valid_rule = self.is_rule_value_valid(
-                    answer_ids_with_parent_id, rule_value
+                    answer_ids_with_parent_id, rule_value, answer_id
```
iwootten: Would be better to modify this to only pass the required answer (answer_ids_with_parent_id[answer_id]["answer"]) to the method.

app/validation/validator.py
```diff
@@ -765,14 +772,13 @@ def _validate_routing_when_dict(self, when_dict, answer_ids_with_parent_id):
...
-        for check_block in answer_ids_with_parent_id:
-            if answer_ids_with_parent_id[check_block]["answer"].get("options"):
-                for answer_block in answer_ids_with_parent_id[check_block][
-                    "answer"
-                ].get("options"):
-                    if rule_value == answer_block.get("value"):
-                        return True
+    def is_rule_value_valid(answer_ids_with_parent_id, rule_value, answer_id):
+        if answer_ids_with_parent_id[answer_id]["answer"].get("options"):
+            for answer_block in answer_ids_with_parent_id[answer_id]["answer"].get(
```
iwootten: This is an option, not an answer_block.

app/validation/validator.py
petechd: Done, thanks!

app/validation/validator.py
petechd: Done, option now, thanks!

app/validation/validator.py
petechd: Passing only answer now, thanks!

app/validation/validator.py
```diff
@@ -729,9 +729,51 @@ def _validate_routing_rule(self, rule, answer_ids_with_group_id, block_or_group)
...
+                    )
 
         return errors
 
+    @staticmethod
+    def is_rule_value_valid(answer, rule_value):
+        for option in answer.get("options", []):
+            if rule_value == option.get("value"):
+                return True
+        return False
```
iwootten:
```suggestion
        return value in option['value'] for option in answer.get("options", [])
```

app/validation/validator.py
```diff
@@ -729,9 +729,51 @@ def _validate_routing_rule(self, rule, answer_ids_with_group_id, block_or_group)
...
+        if when_value:
+            when_values.append(when_value)
+
+        for rule_value in when_values:
+
+            if answer_ids_with_parent_id[answer_id]["answer"].get("type") in (
+                "Radio",
+                "Checkbox",
+            ):
+                answer = answer_ids_with_parent_id[answer_id].get("answer", {})
```
iwootten:
```suggestion
                answer = answer_ids_with_parent_id[answer_id].get("answer", {})
                
                if answer.get("type") in ("Radio", "Checkbox"):
```

app/validation/validator.py
```diff
@@ -2099,8 +2141,19 @@ def _get_answers_with_parent_ids(self, json_to_validate):
         answers = {}
         for question, context in self._get_questions_with_context(json_to_validate):
             for answer in question.get("answers", []):
-                answers[answer["id"]] = {"answer": answer, **context}
+
+                if not answer["id"] in answers:
+                    answers[answer["id"]] = {"answer": answer, **context}
+
                 for option in answer.get("options", []):
+                    if answers[answer["id"]]["answer"]["options"] and option[
```
iwootten:
```suggestion
                    if not self.is_rule_value_valid(answer, value)
                        answer["options"].append(option)
```

app/validation/validator.py
petechd: Thanks!

app/validation/validator.py
```diff
@@ -2099,8 +2135,19 @@ def _get_answers_with_parent_ids(self, json_to_validate):
...
+                        "value"
+                    ] not in [
+                        option["value"]
+                        for option in answers[answer["id"]]["answer"]["options"]
+                    ]:
+                        answers[answer["id"]]["answer"]["options"].append(option)
+
```
iwootten: As discussed, I'm not sure of the need for this section. I can revert this code and none of the tests fail.

app/validation/validator.py
```diff
@@ -729,9 +729,45 @@ def _validate_routing_rule(self, rule, answer_ids_with_group_id, block_or_group)
...
+                    )
+                )
+
+        return errors
+
+    def _validate_routing_when_dict(self, when_dict, answer_ids_with_parent_id):
+        errors = []
+        when_values = when_dict.get("values", [])
+        when_value = when_dict.get("value")
+        answer_id = when_dict.get("id", None)
```
```suggestion
        answer_id = when_dict.get("id")
```

None is the default return value

app/validation/validator.py
iwootten: So it looks like runner validations will fail without this - there should therefore be a test in place within validator that tests this bit of code.

app/validation/validator.py
petechd: Done, thanks!

app/validation/validator.py
petechd: It's a separate method now, thanks! There is invalid schema to test it.

app/validation/validator.py
```diff
@@ -708,21 +725,57 @@ def _validate_routing_rules_default(rules, block_or_group):
...
+                    rule["when"],
+                    answer_ids_with_group_id,
+                    option_value_map,
+                    block_or_group["id"],
                 )
             )
 
         return errors
 
+    def _validate_answer_value_in_when_rule(self, when_rule, option_value_map):
```
iwootten: I think this should be tested through an explicit unit test.

app/validation/validator.py
```diff
@@ -2064,24 +2160,55 @@ def _get_answers_with_parent_ids(self, json_to_validate):
...
+        )
+
+        for answer in radio_and_checkbox_answers:
+            answer_id = answer["id"]
+            option_values = [option["value"] for option in answer["options"]]
+
+            if answer_id in option_value_to_answer_id_map:
+                option_value_to_answer_id_map[answer_id].update(option_values)
+            else:
+                option_value_to_answer_id_map[answer_id] = set(option_values)
```
ajmaddaford: If you change `option_value_to_answer_id_map = {}` to `option_value_to_answer_id_map = defaultdict(set)` then these four lines can be simplified to `option_value_to_answer_id_map[answer_id].update(option_values)`.

app/validation/validator.py
```diff
@@ -2064,24 +2160,55 @@ def _get_answers_with_parent_ids(self, json_to_validate):
...
+        option_value_to_answer_id_map = {}
+        questions = (
+            question
+            for question, _ in cls._get_questions_with_context(json_to_validate)
+        )
+        radio_and_checkbox_answers = cls._get_answers_by_type_from_questions(
+            questions=questions, answer_types={"Radio", "Checkbox"}
+        )
+
+        for answer in radio_and_checkbox_answers:
```
ajmaddaford: Could a lot of this be replaced with a call to `_parse_values(json_to_validate, 'options')` to return all `options` elements?

app/validation/validator.py
```diff
@@ -84,6 +83,9 @@ def validate_questionnaire(self, json_to_validate):
             answers_with_parent_ids = self._get_answers_with_parent_ids(
                 json_to_validate
             )
+            option_value_to_answer_id_map = self._get_option_value_to_answer_id_map(
```
ajmaddaford: As this is a map of answer ids to option values I think it should be named `answer_id_to_option_values_map`.

app/validation/validator.py
```diff
@@ -2064,24 +2160,55 @@ def _get_answers_with_parent_ids(self, json_to_validate):
 
         return answers
 
-    def _get_questions_with_context(self, questionnaire_json):
+    @classmethod
+    def _get_option_value_to_answer_id_map(cls, json_to_validate):
```
iwootten: This seems to be named in reverse? _get_option_value_map seems good enough.

app/validation/validator.py
```diff
@@ -708,21 +740,69 @@ def _validate_routing_rules_default(rules, block_or_group):
...
+        when_values = when_rule.get("values", [])
+        when_value = when_rule.get("value")
+        if when_value:
+            when_values.append(when_value)
+
+        if not Validator._is_when_rule_answer_value_valid(
+            answer_id=when_rule["id"],
+            when_values=when_values,
+            option_value_to_answer_id_map=option_value_to_answer_id_map,
+        ):
```
ajmaddaford: - This should check each value and generate an error for each one that is missing. 
- This could be simplified to:

```python
option_values = option_value_to_answer_id_map.get(when_rule["id"])
for value in when_values:
   if value not in option_values:
       errors.append(
           Validator._error_message(
               f"Answer value in when rule with answer id `{when_rule['id']}` has an invalid value of `{value}`"
           )
       )
```

app/validation/validator.py
MebinAbraham: Nice one.

app/validation/validator.py
MebinAbraham: As discussed, `_parse_values` return a pointer and the object. Therefore, I have made a new method called `get_answers`.

app/validation/validator.py
```diff
@@ -2064,24 +2092,49 @@ def _get_answers_with_parent_ids(self, json_to_validate):
 
         return answers
 
-    def _get_questions_with_context(self, questionnaire_json):
+    @classmethod
+    def _get_answer_id_to_option_values_map(cls, json_to_validate):
```
iwootten: We should probably make this a class method and use using self.schema?

app/validation/validator.py
```diff
@@ -2064,24 +2092,49 @@ def _get_answers_with_parent_ids(self, json_to_validate):
...
+
+    @classmethod
+    def _get_answers(cls, json_to_validate):
+        questions = cls._get_questions_with_context(json_to_validate)
+        for question, _ in questions:
+            for answer in question["answers"]:
+                yield answer
+
+    @classmethod
+    def _get_questions_with_context(cls, questionnaire_json):
```
iwootten: Same as above

app/validation/validator.py
MebinAbraham: I don't understand your comment. This is already a class method. `self.schema` isn't same as `json_to_validate`. Can you elaborate on what exactly you mean?

iwootten: I mean an instance method, self.schema is already held so it seems redundant to expect it to be passed.

MebinAbraham: Unless I misunderstood, `self.schema` is `questionnaire_v1.json` and not the actual schema that is to be validated?

iwootten: I see, as you were then. Stand down.

PR #176, opened 2019-12-03
Propose plural forms in schema text

ajmaddaford:
### PR Context

See https://github.com/ONSdigital/eq-schema-validator/blob/propose-plural-forms-in-schema-text/doc/decisions/0003-plural-forms-in-schema-text.md for the proposal to add plural forms to the schema. 

This pull request also adds the required schema. 

The code to check for valid value sources has been factored out of the placeholders validation. It now searches the schema for all references to value sources and checks their validity against answer ids, metadata and lists.

### DIFF
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -1722,6 +1722,8 @@ class Validator:  # pylint: disable=too-many-lines
         elif isinstance(input_data, list):
             for item in input_data:
                 yield from self._get_dicts_with_key(item, key_name)
+        else:
+            yield from ()
 
     def _validate_placeholder_object(self, placeholder_object, current_block_id):
         """ Current block id may be None if called outside of a block
@@ -1766,7 +1768,7 @@ class Validator:  # pylint: disable=too-many-lines
     def _validate_placeholders(self, block_json):
         errors = []
         strings_with_placeholders = self._get_dicts_with_key(block_json, "placeholders")
-        for placeholder_object in strings_with_placeholders or []:
+        for placeholder_object in strings_with_placeholders:
             errors.extend(
                 self._validate_placeholder_object(placeholder_object, block_json["id"])
             )
@@ -1778,7 +1780,7 @@ class Validator:  # pylint: disable=too-many-lines
     ):
         errors = []
         source_references = self._get_dicts_with_key(block_json, "identifier")
-        for source_reference in source_references or []:
+        for source_reference in source_references:
             source = source_reference["source"]
             if isinstance(source_reference["identifier"], str):
                 identifiers = [source_reference["identifier"]]
--- a/app/validation/validator.py
+++ b/app/validation/validator.py
@@ -1817,15 +1817,14 @@ class Validator:  # pylint: disable=too-many-lines
                         f"Invalid answer reference '{identifier}' in block '{current_block_id}'"
                     )
                 )
-            else:
-                if answers_with_parent_ids[identifier]["block"] == current_block_id:
-                    errors.append(
-                        self._error_message(
-                            "Invalid answer reference '{}' in block '{}' (self-reference)".format(
-                                identifier, current_block_id
-                            )
+            elif answers_with_parent_ids[identifier]["block"] == current_block_id:
+                errors.append(
+                    self._error_message(
+                        "Invalid answer reference '{}' in block '{}' (self-reference)".format(
+                            identifier, current_block_id
                         )
                     )
+                )
         return errors
 
     def _validate_metadata_source_reference(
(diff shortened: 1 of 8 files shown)


app/validation/validator.py
```diff
@@ -1734,185 +1723,149 @@ def _get_dicts_with_key(self, input_data, key_name):
...
-                            answer_id_to_validate, placeholder_definition["placeholder"]
-                        )
+                        f"Invalid answer reference '{identifier}' in block '{current_block_id}'"
                     )
                 )
-                continue
-
-            answer_block_id = answers_with_parent_ids[answer_id_to_validate]["block"]
+            else:
+                if answers_with_parent_ids[identifier]["block"] == current_block_id:
```
MebinAbraham: This could be?
```py
elif answers_with_parent_ids[identifier]["block"] == current_block_id:
```

schemas/string_interpolation/definitions.json
```diff
@@ -152,5 +166,42 @@
...
     "required": ["source", "identifier"]
+  },
+  "text_plural": {
+    "description": "A mapping of a count to plural strings",
+    "type": "object",
+    "properties": {
+      "forms": {
+        "description": "Plural forms of the text. Uses the CLDR plural categories. One or more of these properties will be present dependent on the language.",
+        "type": "object",
+        "properties": {
```
MebinAbraham: These should be under `"items"` with a ` "minItems": 2`.

app/validation/validator.py
```diff
@@ -1734,185 +1723,149 @@ def _get_dicts_with_key(self, input_data, key_name):
...
-    def _validate_relationship_collector_answers(self, answers):
-        one_answer_msg = "RelationshipCollector contains more than one answer."
-        answer_type_msg = "Only answers of type Relationship are valid in RelationshipCollector blocks."
-
+    def _validate_source_references(
+        self, block_json, answers_with_parent_ids, valid_metadata_ids
+    ):
         errors = []
+        source_references = self._get_dicts_with_key(block_json, "identifier")
+        for source_reference in source_references or []:
```
JamesGardiner: if self._get_dict_with_key returned an empty iterable by default then the `or []` pattern which is used frequently wouldn't be needed

schemas/string_interpolation/definitions.json
ajmaddaford: It's not an array so no need for items, but the `additionalProperties` and `required` properties are at the wrong level so I'll move them.

schemas/string_interpolation/definitions.json
```diff
@@ -152,5 +166,42 @@
...
+          },
+          "few": {
+            "type": "string"
+          },
+          "many": {
+            "type": "string"
+          },
+          "other": {
+            "type": "string"
+          },
```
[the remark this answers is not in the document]
ajmaddaford: No, because that would allow an object with a `placeholders` within each plural form i.e.

```json
{
    "summary_text": {
        "placeholders": [
            {
                "placeholder": "number_of_people",
                "transforms": []
            }
        ],
        "text_plural": {
            "forms": {
                "one": "Yes, {number_of_people} person lives here",
                "other": {
					"text": "Yes, {number_of_people} people live here",
					"placeholders": [
						{}
					]
				}
            },
            "count": {
                "source": "list",
                "identifier": "household"
            }
        }
    }
```

The intent is that the `text_plural` element is resolved to a string equivalent of a `text` element, which is then combined with the resolved placeholders. If there were more placeholders than the number in the sentence then it would look something like:

```json
{
    "summary_text": {
        "placeholders": [
            {
                "placeholder": "number_of_people",
                "transforms": [
                {
                    "arguments": {
                        "number": {
                            "source": "list",
                            "identifier": "household"
                        }
                    },
                    "transform": "number_to_words"
                }
                ]
			{
				"placeholder": "address",
                "value": {
					"source": "metadata",
					"identifier": "display_address"
				}
            }
        ],
        "text_plural": {
            "forms": {
                "one": "Yes, {number_of_people} person live at {address}",
                "other": "Yes, {number_of_people} people live at {address}"
            },
            "count": {
                "source": "list",
                "identifier": "household"
            }
        }
    }
```

I'll add this as an example to the document.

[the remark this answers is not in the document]
MebinAbraham: Ah ay, I confused myself because of the parenthesis.

schemas/string_interpolation/definitions.json
```diff
@@ -152,5 +166,42 @@
...
+          }
+        },
+        "additionalProperties": false,
+        "required": ["other"]
+      },
+      "count": {
+        "description": "The source to use for the count. Must resolve to a number.",
+        "$ref": "#/lookup_source"
+      },
+      "additionalProperties": false
```
MebinAbraham: This is nested 1 too deep.

schemas/string_interpolation/definitions.json
```diff
@@ -152,5 +166,42 @@
...
+    }
```
MebinAbraham: After this, we need to define `"required": ["forms", "count"]`

PR #177, opened 2019-12-05
Add the cancel_text variable to the question schema

berroar:
This change is required to support the eq-questionnaire-runner functionality that has been added [here](https://github.com/ONSdigital/eq-questionnaire-runner/pull/13).

I have added the new 'cancel_text' variable to the question schema.

### DIFF
--- a/schemas/blocks/list_add_edit_remove_question.json
+++ b/schemas/blocks/list_add_edit_remove_question.json
@@ -15,11 +15,7 @@
       },
       "type": {
         "type": "string",
-        "enum": [
-          "ListAddQuestion",
-          "ListEditQuestion",
-          "ListRemoveQuestion"
-        ]
+        "enum": ["ListAddQuestion", "ListEditQuestion", "ListRemoveQuestion"]
       },
       "show_on_section_summary": {
         "type": "boolean"
--- a/schemas/blocks/list_add_edit_remove_question.json
+++ b/schemas/blocks/list_add_edit_remove_question.json
@@ -1,5 +1,5 @@
 {
-  "$id": "https://eq.ons.gov.uk/blocks/question.json",
+  "$id": "https://eq.ons.gov.uk/blocks/list_add_edit_remove_question.json",
   "$schema": "http://json-schema.org/draft-07/schema",
   "block": {
     "type": "object",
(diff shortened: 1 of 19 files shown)


berroar:
[Ability to cancel from add a person screen (S)](https://trello.com/c/VOlzRmlO/3533-ability-to-cancel-from-add-a-person-screen-s)

berroar:
[Ability to cancel from add a person screen (S)](https://trello.com/c/VOlzRmlO/3533-ability-to-cancel-from-add-a-person-screen-s)

schemas/blocks/question.json
```diff
@@ -10,6 +10,9 @@
       "number": {
         "type": "string"
       },
+      "cancel_text": {
+        "type": "string"
+      },
```
MebinAbraham: This would technically mean all blocks would support `cancel_text` and not just the `add_block`. Not sure if it worth making an add_block question type just for it.

ajmaddaford: I'm not sure why it should just be on the add block. Maybe it should be on add, change and remove?

berroar: I've added a new question type for list add/edit/remove questions.

schemas/blocks/list_add_edit_remove_question.json
```diff
@@ -0,0 +1,69 @@
...
+      },
+      "type": {
+        "type": "string",
+        "enum": [
+          "ListAddQuestion",
+          "ListEditQuestion",
+          "ListRemoveQuestion"
+        ]
+      },
+      "show_on_section_summary": {
```
ajmaddaford: This doesn't apply to list add/edit/remove questions.

schemas/blocks/list_add_edit_remove_question.json
```diff
@@ -0,0 +1,69 @@
...
+        "type": "boolean"
+      },
+      "question": {
+        "$ref": "https://eq.ons.gov.uk/questions/definitions.json#/question"
+      },
+      "question_variants": {
+        "$ref": "https://eq.ons.gov.uk/blocks/definitions.json#/question_variants"
+      },
+      "list_summary": {
```
ajmaddaford: This should be broken out into a separate definition that can be referenced by this and the original question block json.

PR #180, opened 2020-01-08
Add visible write-ins

petechd:
### PR Context
This adds new boolean "open" property to be used in some voluntary radio and checkbox answers. It needs to be merged before changes described on [this Trello card](https://trello.com/c/g3esxHDn/3433-visible-write-ins-m) can be implemented. [PR of the runner change](hhttps://github.com/ONSdigital/eq-questionnaire-runner/pull/25).

### DIFF
--- a/schemas/answers/text_field.json
+++ b/schemas/answers/text_field.json
@@ -26,6 +26,9 @@
       "mandatory": {
         "type": "boolean"
       },
+      "open": {
+        "type": "boolean"
+      },
       "validation": {
         "type": "object",
         "properties": {
--- a/schemas/answers/text_field.json
+++ b/schemas/answers/text_field.json
@@ -26,7 +26,7 @@
       "mandatory": {
         "type": "boolean"
       },
-      "open": {
+      "visible": {
         "type": "boolean"
       },
       "validation": {
--- a/schemas/answers/number.json
+++ b/schemas/answers/number.json
@@ -26,6 +26,9 @@
       "mandatory": {
         "type": "boolean"
       },
+      "visible": {
+        "type": "boolean"
+      },
       "max_length": {
         "type": "integer"
       },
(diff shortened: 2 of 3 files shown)


petechd:
[visible write ins (M)](https://trello.com/c/g3esxHDn/3433-visible-write-ins-m)

schemas/answers/text_field.json
```diff
@@ -26,6 +26,9 @@
       "mandatory": {
         "type": "boolean"
       },
+      "open": {
```
MebinAbraham: I personally think it is clearer for this to be `visible` or `is_visible`. Thoughts?

JamesGardiner: visible - we already specify that it is a boolean so the `is_` seems superfluous.

petechd: Ok, sorted now, thanks!

petechd: The only thing is: "visible" is used in schemas sometimes(in obsolete "navigation"), but I discussed it with the guys and they gave me the green light.

schemas/answers/text_field.json
```diff
@@ -26,6 +26,9 @@
       "mandatory": {
         "type": "boolean"
       },
+      "visible": {
```
MebinAbraham: It would be good to add this to an existing valid schema under `tests/`

MebinAbraham: This key should be part of detail answers and not text field?

MebinAbraham: Should also be added to numbers.json as details answer can be a text field or a number.

petechd: I've added one to `test_placeholder_source_ids.json`. It passes fine. Thanks!

petechd: Added, thanks.

PR #182, opened 2020-01-20
Propose routing rules improvements

ajmaddaford:
### PR Context

There are some issues with the current structure of routing rules:

- They only support limited `AND` and `OR` logic - `OR` is only supported outside of a `when` block leading to duplicate routing targets 
- Logic can't be nested for more nuanced rules
- Dynamic values (answers, metadata, lists) are referenced inconsistently

This pull request proposes a new structure to solve these issues.

### DIFF
--- a/doc/decisions/0004-extend-and-improve-routing-rules.md
+++ b/doc/decisions/0004-extend-and-improve-routing-rules.md
@@ -53,7 +53,7 @@ The rule definition is:
 
 ```json
 {
-	"and": [
+    "and": [
         {
             "or": [
                 {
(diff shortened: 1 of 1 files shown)


ajmaddaford:
@MebinAbraham I agree about `id_selector`, the conversions are examples of how the rules could look; the actual implementation can be decided when we make the changes.

ajmaddaford:
@LJBabbage For me, an expression is a combination of a value and an operator. So in our case an expression is an operation and its arguments. `argument` was chosen over `value` to indicate that the value can be unknown. Also, I'd consider this in mathematics terms as a function rather than an equation as the arguments are assigned rather than calculated (http://www.differencebetween.net/language/words-language/difference-between-equations-and-functions/).

